PR #6, opened 2023-08-15
Feature/data table

buckhalt:
Port `DataTable` Component for protocol, session, and participant management. Component includes:

- Row actions dropdown
- Column sorting
- Row selection
- Automatic pagination in pages of 10. Data is not paginated server-side. This is implemented for improved usability, not for optimization.
- Filter using search input
- Delete Selected button which is enabled when rows are selected

This PR also includes seeding for Protocols, Sessions, and Participants using Faker.

CU-8685ax4wd


jthrilly:
Task linked: [CU-8685ax4wd Port data table component](https://app.clickup.com/t/8685ax4wd)

components/ui/DataTable/Columns.tsx
```diff
@@ -0,0 +1,222 @@
+'use client';
+
+import { type ColumnDef } from '@tanstack/react-table';
+import { type Interview, type Protocol, type User } from '@prisma/client';
```
jthrilly, line 4: Where everything is a type, you can just put type before the named imports:

```suggestion
import type { Interview, Protocol, User } from '@prisma/client';
```

app/(main)/page.tsx
```diff
@@ -41,6 +60,7 @@ export default async function Home() {
             <InterviewCard key={interview.id} interview={interview} />
           ))}
         </div>
+        <DataTable columns={InterviewColumns} data={interviews} />
```
jthrilly, line 63: I know this implementation is really just an example, but I think the ideal way of implementing this would be:

- Add the `<DataTable />` to the root `components` folder.
- Create a `<InterviewsTable />` component in the _route folder_ (assuming we anticipate it will only be used on a single route. So in this case, putting it in `app/(main)/_components`. 
- Ideally, putting the component in its own subfolder too, which would allow the interview specific column definitions in `Columns.tsx` to be moved there, too.
- Not sure if this is possible, but providing a _default_ columns definition that will be automatically generated based on whatever the structure of the data parameter is. So obviously nothing fancy like sorting or selecting - just rendering a table.

app/(main)/page.tsx
```diff
@@ -41,6 +60,7 @@ export default async function Home() {
             <InterviewCard key={interview.id} interview={interview} />
```
jthrilly, line 60: Let's remove the existing card rendering that I added as a placeholder.

components/ui/DataTable/ActionsDropdown.tsx
```diff
@@ -0,0 +1,34 @@
...
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuLabel,
+  DropdownMenuTrigger,
+} from '~/components/ui/dropdown-menu';
+import { MoreHorizontal } from 'lucide-react';
+import { Button } from '~/components/ui/Button';
+
+type MenuItem = 'Edit' | 'Delete' | 'Resume';
```
jthrilly, line 11: It would be good if this could accept a generic so that the actions could be dynamically defined depending on the use-case.

jthrilly (changes requested):
Looking really good! Have a read of my feedback, and let me know if you have questions.

app/(main)/page.tsx
jthrilly: And then doing the same for `<ProtocolsTable />`, `<ParticipantsTable />` etc...

buckhalt: Resolved with [480c1c1](https://github.com/complexdatacollective/Fresco/pull/6/commits/480c1c11df0dcead08587a36bd746ddbc3f73c7c), [4ee0915](https://github.com/complexdatacollective/Fresco/pull/6/commits/4ee091543ee1b8fa2f1faf445f0ff9e9c2b3f6ac), [22c526c](https://github.com/complexdatacollective/Fresco/pull/6/commits/22c526c1bf50733989da048bc9145da3a6ac0fe1)

components/ui/DataTable/ActionsDropdown.tsx
buckhalt: resolved with [fc91b0e](https://github.com/complexdatacollective/Fresco/pull/6/commits/fc91b0e8bb86177ae4d5927955474280dd03cc75)

jthrilly (approved):
Looks good to me! 

Discussed date/time rehydration and decided it was OOS for this issue - to be looked at later.

Discussed security issue and will work on that next.

PR #8, opened 2023-08-30
Data mapper

jthrilly:
Purpose of this PR is to implement a wrapper around prisma queries so that they are "safe", which means that they do not return unexpected data.

Motivation for this is that it is easy to accidentally pass sensitive info to client javascrtipt bundles in nextjs by passing the result of a prisma query into a client component. 

This PR also implements ES lint rule enforcement to make sure that the new pattern is followed.

Authored by @buckhalt.


buckhalt:
This PR implements a data mapper, `safeLoader`, using `zod` for validation. This helps prevent security issues that come from fetching indiscriminately in Prisma.

This pattern is enforced with a custom eslint rule which requires instances of `prisma.*` to occur within a `safeLoader`. 


#CU-8685dzfhc

jabu404 (changes requested):
Looks good to me, can we remove console.logs, though?

mrkarimoff (commented):
It all looks good to me

jthrilly:
Just a reminder to please use branch names like `feature/data-mapper`, `fix/horrible-bug`.

PR #9, opened 2023-09-01
Make db-seed script cross-platform compatibile

buckhalt:
`"db-seed": "NODE_ENV=development prisma db seed"` does not work on windows because windows requires using SET in the command to set environment variables inline. 

`cross-env` solves this by allowing for setting and using environment variables inline across platforms.


jthrilly (approved):
Seems like `cross-env` is still perfectly okay to use, so I say we merge this. Thanks Caden!

PR #11, opened 2023-09-20
Feature/participant management

mrkarimoff:
#### Created the following features:
- Ability to view all study participants using a data table.
- Ability to add a single participant via a modal dialog or similar
- Ability to import CSV of participants
- Ability to export CSV of participants


jthrilly (approved):
We're going to merge this in now, and then keep working on the final features/refinements in separate branches.

PR #15, opened 2023-09-22
Feature/active protocol

buckhalt:
### Implement the concept of currently 'active' protocol

When a new protocol is added via the ProtocolUploader, the user can decide if they want to mark it as active.
If marked as active, the other protocols will be marked as inactive.

Screenshot 2023-09-22 at 12 39 35 PM


mrkarimoff (approved):
I didn't find any problems in your code. This looks good

PR #16, opened 2023-09-26
Feature/interview management view

buckhalt:
- implements interview management view which allows researchers to see interviews that have been conducted
- this includes information about the interview, participant, and protocol that was used


jthrilly (approved):
I think this looks fine as a first pass. 

Sooner or later we will want to do something about the UX of deleting interviews/protocols/participants. Right now there's very little user feedback about what is going on. That can wait for later, though.

It would also be good to update the table to use richer components for some of the columns. For example, "step" should probably be replaced by "progress", which should in fact be a progressbar. Similarly, the participant fields should link to the participant in the participant route (something like `participant/identifier` - of course this doesn't exist at present).

PR #20, opened 2023-10-06
Feature/anonymous recruitment

buckhalt:
This PR implements optional anonymous recruitment setting. This feature includes:

- Switch component on onboarding and on dashboard to set setting
- If anonymous recruitment is enabled and there is an active protocol, route `/interview/new` is enabled. 
- This route accepts an optional query string `?identifier=`. If provided, it uses that to create a new interview and participant with that identifier. If none is provided, a random UUID is created to be used as the identifier.
- After the interview is created, it reroutes to` /interview/[id]` of newly created interview


app/(interview)/interview/new/page.tsx
```diff
@@ -3,57 +3,71 @@
...
-  // onboarding flow.
-  const session = false;
-  console.error('UPDATE ME');
+  // Create a new interview
+  const identifier = searchParams.identifier || faker.string.uuid();
+  const { createdInterview } = await trpc.interview.create.mutate(identifier);
 
-  if (!session) {
-    redirect('/');
+  if (!createdInterview) {
```
jthrilly, line 66: I think we need to handle the case that the identifier already exists. It seems like we might need to separate out the case where the identifier is provided in the searchParams and where it is generated. 

For situations where it is provided, we need to show a message that the identifier is in use. You should be able to update the return type of the mutation to have a property that can indicate this.

prisma/schema.prisma
```diff
@@ -94,6 +94,7 @@ model SetupMetadata {
   configured    Boolean   @default(false)
   initializedAt DateTime  @default(now())
   configuredAt  DateTime?
+  allowAnonymousRecruitment Boolean   @default(false)
```
jthrilly, line 97: Just wondering if this model should now be called something like "AppSettings" rather than SetupMetadata. It seems to make more sense.

server/routers/interview.ts
```diff
@@ -1,55 +1,80 @@
...
+      }
     }),
-    protocol: z.object({
-      id: z.string(),
-      name: z.string(),
+  get: router({
+    all: publicProcedure.query(async () => {
+      const interviews = await prisma.interview.findMany({
+        include: {
+          protocol: true,
```
jthrilly, line 59: I think this also needs to include the participant, since this is used primarily by the interview dashboard route to render the table. 

We should think about if it makes sense to configure what is included in the query input. 

Actually on that point, we need to think about pagination and whatnot with these large queries. I'll add a clickup task to cover that.

server/routers/interview.ts
```diff
@@ -1,55 +1,80 @@
...
-      loader: () =>
-        prisma.interview.findMany({
+    byId: publicProcedure
+      .input(interviewIdSchema)
+      .query(async ({ input: id }) => {
+        const interview = await prisma.interview.findFirst({
+          where: id,
           include: {
             protocol: true,
-            participant: true,
```
jthrilly, line 45: Curious about why this was removed. Can you say more?

shared/schemas.ts
```diff
@@ -11,3 +11,9 @@ export const updateSchema = z.object({
   identifier: participantIdentifierSchema,
   newIdentifier: participantIdentifierSchema,
 });
+
+export const interviewIdSchema = z.object({
+  id: z.string(),
```
jthrilly, line 16: I'm not sure this is worth a shared schema. Can't we infer from the prisma model?

jthrilly (changes requested):
Looking awesome! Functionality is really nice. Just a few thoughts from me, which you can choose how you want to handle. Let me know if you want to talk through anything else.

Take note of the changes I made in my few commits and see my commit message for details.

app/(interview)/interview/new/page.tsx
jthrilly: Just while I think about it, there are other cases too: identifier isn't provided, identifier isn't valid (too long, short).

server/routers/interview.ts
buckhalt: included participant here [0ea46cc](https://github.com/complexdatacollective/Fresco/pull/20/commits/0ea46cca62dc0de59eeb44ff7b3264c910a23955)

server/routers/interview.ts
buckhalt: resolved with [0ea46cc](https://github.com/complexdatacollective/Fresco/pull/20/commits/0ea46cca62dc0de59eeb44ff7b3264c910a23955)

app/(interview)/interview/new/page.tsx
buckhalt: resolved [ec1f4f3](https://github.com/complexdatacollective/Fresco/pull/20/commits/ec1f4f378d47d5b2b104a4c2d69fbbdafc55c6b8)

shared/schemas.ts
buckhalt: resolved [400e1ec](https://github.com/complexdatacollective/Fresco/pull/20/commits/400e1ec3c3aa14a9f73e75a234a2aa3c9cb4fd94). 

Decided to move schema directly into input because deriving zod schema from existing types is not yet supported.

jthrilly (approved):
Hot.

PR #24, opened 2023-10-12
Feature/network context provider

jthrilly:
New implementation of this, which switches from using a route parameter to a search parameter for the stage. 

Ignore the build/ts errors and just check out the implementation for any obvious errors or places to improve.


mrkarimoff (approved):
This all looks great. A little overwhelming for me, but great

providers/InterviewProvider.tsx
```diff
@@ -0,0 +1,124 @@
...
+}) {
+  const [currentStage, setCurrentStage] = useQueryState(
+    'stage',
+    parseAsInteger.withDefault(1),
+  );
+  const [initialized, setInitialized] = useState(false);
+  const { network, networkHandlers } = useNetwork(initialNetwork);
+  const { mutate: updateNetwork } = trpc.interview.updateNetwork.useMutation();
+
+  const stages = JSON.parse(protocol?.stages) as Stage[];
```
buckhalt, line 55: protocol.stages is unexpectedly typed as a string, causing issues with length and accessing stageConfig by index. This workaround fixes those. Now, the correct stage data is rendered, and the navigation length is correct.

We may want to consider looking into why the stages are being serialized as a string initially though. Any ideas?

buckhalt (approved):
Really nice job getting this to work! See my commit and comment for a type issue I found with stages, but otherwise looking great.

providers/InterviewProvider.tsx
jthrilly: That's strange...

I think this could be another one caused by prisma schema issues on your side. The `protocol` prop isn't JSON - it is of type `Protocol` (coming from shared consts). It doesn't need to be parsed to be used directly - its already an object with specific known keys. 

If this code was failing on your side it could be because the schema was updated to store the protocol as JSON, which then comes back from the database call which fetches the interview as `Prismma.JsonValue`. As I say, this doesn't needed to be parsed into a javascript object, because JSON objects are automatically converted by prisma. I'm forcing it to be typed as a `Protocol` in the page handler because I think its reasonable to assume it will always be a protocol.

So based on that, I've reverted this change. I'll see if I come across any issues.

PR #25, opened 2023-10-12
Feature/participant deletion

buckhalt:
### Features
- Adds prompt for deleting participants
- Prompt has warning if participant has interview data
- Prompt has more serious warning if participant has un-exported interview data
- Refactor delete all participants functionality to use these prompts

### UI:

Deleting participants which have no interview data:
Screenshot 2023-10-12 at 12 02 23 PM

Deleting participants with exported interview data:
Screenshot 2023-10-12 at 12 03 10 PM

Deleting participants with un-exported interview data
Screenshot 2023-10-12 at 12 32 25 PM


app/(dashboard)/dashboard/_components/ParticipantsTable/ParticipantsTable.tsx
```diff
@@ -33,17 +39,40 @@ export const ParticipantsTable = ({
...
+    if (deleteAll) {
+      await deleteAllParticipants();
+      await refetch();
+      setShowAlertDialog(false);
+    }
+
+    // Delete selected participants
+    await deleteParticipants(participantsToDelete.map((d) => d.identifier));
     await refetch();
+    setShowAlertDialog(false);
```
mrkarimoff, line 75: I noticed here `setShowAlertDialog(false)` is being invoked in either the condition true or false, so maybe we don't need one in line 69.

buckhalt: resolved with [ba9fa0e](https://github.com/complexdatacollective/Fresco/pull/25/commits/ba9fa0eff2eaf68da29e6d9373ab3bd9e1233bee)

mrkarimoff (approved):
The changes look good!

jthrilly (changes requested):
Hmm...I don't really like how this is developing! I see a number of issues:

- We've got state items for two different modals, with confusing/ambiguous names (see my other review comment).
- We've got state for editing items, which is set indirectly by passing a function that sets it into another component (not implemented here, but still a bad design).
- We've got derived properties being calculated on each render within a component (such as `hasInterviews`) rather than being calculated by the parent component and passed as a simpler prop.
- We've got confusing naming. `hasInterviewsNotYetExported`! How about `hasUnexportedInterviews`?!
- We've got a state item that determines behaviour (delete single vs all) in a really confusing way. 

What most of these ultimately come down to is that we have got a really poor pattern with regards to the way that the data table interacts with actions that can be performed. This leads to us having to pass state around.

We want to get to a place where we can: 

1.  define dialogs once, but use them in multiple places. Right now, we have a single dialog, but we have to pass its state around through different components. Its really ugly and not scalable. 
2. have our table actions be able to trigger their own dialogs to confirm or prompt as they see fit. ideally, these dialogs will be instances of the shared dialogs in point 1. 

I think the best way to achieve this is to refactor the way the actions dropdown works in the data table, so that instead of defining dropdown items, you provide the whole dropdown. This is somewhat complicated by the fact that triggering a dialog within a dropdown with shadcn and radix requires a workaround (see the note here: https://ui.shadcn.com/docs/components/dialog#notes) but it can be done if the whole component is defined together. There's an example of how this should work here: https://codesandbox.io/embed/r9sq1q (specifically, `DropdownWithDialogItemsSolution2`).

Basically, the DataTable should accept an optional `actions` prop, which contains a `DropdownMenu` (just like the current `ActionsDropdown`. If provided, this will be rendered automatically as an additional column (so that column definitions no longer need to include the actions. For this to work, we will need to pass the current row item to the actions dropdown so that actions can do things with the item they reference.

Once this is done, the implementation for this feature can be re-written as:

1. A confirm dialog component, with a simpler props interface. Something like:
```
interface DeleteParticipantConfirmationDialog {
  open: boolean;
  onConfirm: () => Promise<void>;
  onCancel: () => void;
  numberOfParticipants?: number;
  hasInterviews?: boolean;
  hasUnexportedInterviews?: boolean;
  isDeleting?: boolean;
}
```
2. A "delete all" button component, which has its own dialog state which it uses to render the dialog. This is rendered in `ParticipantsTable`.
3. A custom `ActionsDropdown` for the `ParticipantsTable` that has dialog state for deleting and editing dialogs (using `ParticipantModal` and the shared delete modal).
4. An "add participant" button component that has its own dialog state which it uses to render a `ParticipantModal`. Currently, creating a new participant and editing a participant uses the same dialog.

app/(dashboard)/dashboard/_components/ParticipantsTable/ParticipantsTable.tsx
```diff
@@ -1,24 +1,30 @@
...
   initialData,
 }: {
-  initialData: Participant[];
+  initialData: ParticipantWithInterviews[];
 }) => {
   const [seletedParticipant, setSeletedParticipant] = useState<string | null>(
     null,
   );
   const [showModal, setShowModal] = useState(false);
+  const [showAlertDialog, setShowAlertDialog] = useState(false);
```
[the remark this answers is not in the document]
buckhalt: resolved with [620724f](https://github.com/complexdatacollective/Fresco/pull/25/commits/620724f90bdc4f5cafc8c5f78e0c4cc58290f02b)

app/(dashboard)/dashboard/_components/ParticipantsTable/ParticipantsTable.tsx
```diff
@@ -1,71 +1,70 @@
...
         handleDeleteSelected={handleDelete}
+        actions={[
+          {
+            id: 'actions',
+            header: () => <Settings />,
+            cell: ({ row }) => {
+              return <ActionsDropdown row={row} participants={participants} />;
+            },
+          },
+        ]}
```
jthrilly, line 67: I think a better pattern for this would be:

```
<DataTable
  actions={ActionsDropdown}
  actionRowHeader={Settings}
/>
```

Although its convenient to pass in the structure exactly as you need to consume it in data table, it doesn't really make sense as a prop. Remember that when we expose props we are thinking of the consumer's interface for the component. The consumer doesn't know or care how the component works internally, and usually is most concerned with being able to quickly understand the props interface. Doing it the way I suggested will give us that, while also giving is simpler typing.

A couple of things to point out with your approach:
  - you're passing participants in to the actions dropdown. I would suggest making it so that whatever component is passed as the actions component is passed the row and the data as properties by the data table, since it is already being given the data prop itself. This simplifies things, and should be able to be typed so that the component you pass must accept those props (though be careful with this - maybe they are both optional?). 
  - its a good idea to separate out props that are optional from ones that are required. For example, `actionRowHeader` can default to something sensible (although I suggest maybe making it blank by default, as I did find the settings cog confusing) and therefore be an optional property. 

Hope its obvious enough how to make this work inside data table, but let me know if not.

jthrilly (changes requested):
Looking good so far! See what you think about my note.

app/(dashboard)/dashboard/_components/ParticipantsTable/ParticipantsTable.tsx
buckhalt: resolved with [13bdd5a](https://github.com/complexdatacollective/Fresco/pull/25/commits/13bdd5ae29b795c64c7a197bdf40db65c8eabc69)

I made `actionsHeader` blank by default, and opted to leave it as blank for both of the tables for now. If we want to add something back in for that, we can.

components/DataTable/DataTable.tsx
```diff
@@ -28,7 +29,8 @@ interface DataTableProps<TData, TValue> {
   data: TData[];
   filterColumnAccessorKey?: string;
   handleDeleteSelected?: (data: TData[]) => Promise<void> | void;
-  actions?: ColumnDef<TData, TValue>[];
+  actions?: React.ComponentType<{ row: Row<TData>; data: TData[] }>;
```
jthrilly, line 32: Nice! Didn't know this is how you did it. This is super cool, because it also validates the data type being passed to the table!

app/(dashboard)/dashboard/_components/ProtocolsTable/ProtocolsTable.tsx
```diff
@@ -50,15 +49,7 @@ export const ProtocolsTable = ({
...
-        actions={[
-          {
-            id: 'actions',
-            header: () => <Settings />,
-            cell: ({ row }) => {
-              return <ActionsDropdown row={row} />;
-            },
-          },
-        ]}
+        actions={ActionsDropdown}
```
jthrilly, line 52: The component API for data table is looking great now!

jthrilly (approved):
Looks great to me! Hope you agree that the refactor was useful.

PR #27, opened 2023-10-16
Feature/onboarding upload participants

buckhalt:
Updates onboarding process to use new upload participants/csv processing code

### DIFF
--- a/app/(dashboard)/dashboard/participants/_components/ImportCSVModal.tsx
+++ b/app/(dashboard)/dashboard/participants/_components/ImportCSVModal.tsx
@@ -127,9 +127,10 @@ const ImportCSVModal = ({ onUploaded }: { onUploaded?: () => void }) => {
         description: 'Participants have been imported successfully',
         variant: 'success',
       });
-      if (typeof onUploaded === 'function') {
-        onUploaded();
-      }
+    }
+
+    if (typeof onUploaded === 'function') {
+      onUploaded();
     }
 
     methods.reset();
--- a/app/(dashboard)/dashboard/participants/_components/ImportCSVModal.tsx
+++ b/app/(dashboard)/dashboard/participants/_components/ImportCSVModal.tsx
@@ -25,7 +25,11 @@ const formSchema = z.object({
   csvColumn: z.string().optional(),
 });
 
-const ImportCSVModal = ({ onUploaded }: { onUploaded?: () => void }) => {
+const ImportCSVModal = ({
+  onImportComplete,
+}: {
+  onImportComplete?: () => void;
+}) => {
   const { toast } = useToast();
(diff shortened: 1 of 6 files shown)


mrkarimoff (approved):
This all looks good to me. I don't see anything that needs to be changed 👍

app/(onboard)/layout.tsx
```diff
@@ -19,6 +20,7 @@ export default function Layout({ children }: PropsWithChildren) {
             />
           </Link>
         </div>
+        <Toaster />
```
jthrilly, line 23: This means we have two of this component in separate layouts. Because there's already a `Toaster` in the dashboard layout, you should instead add this to the root layout, and remove it from here and the dashboard layout.

app/(dashboard)/dashboard/participants/_components/ImportCSVModal.tsx
```diff
@@ -129,6 +129,10 @@ const ImportCSVModal = () => {
       });
     }
 
+    if (typeof onUploaded === 'function') {
+      onUploaded();
+    }
+
```
jthrilly, line 135: 🤨 

- The component's verb is _import_, but you've decided to call the callback "uploaded". Let's make it `onImportComplete` or something.
- You made the callback optional with a generic function signature when it is defined. Optional properties are `null` by default, but you are instead testing if it is a function. You can just do `if (onImportComplete)`, or better yet you can use optional chaining and just call it if it exists: `onImportComplete?.()`.

jthrilly (changes requested):
I might be going crazy here, but wasn't this feature supposed to be about using the new upload _protocol_ functionality? Doesn't this just add CSV upload?

buckhalt:
> I might be going crazy here, but wasn't this feature supposed to be about using the new upload _protocol_ functionality? Doesn't this just add CSV upload?

Sorry, I have no idea why I named this PR incorrectly. Fixed the name now. This is supposed to be https://app.clickup.com/t/8685ynge0, adding the csv participants upload.

app/(onboard)/layout.tsx
buckhalt: resolved with [ea02035](https://github.com/complexdatacollective/Fresco/pull/27/commits/ea020353c797666bf77a7abd67e6ff446c452ac9)

app/(dashboard)/dashboard/participants/_components/ImportCSVModal.tsx
buckhalt: resolved here [4060004](https://github.com/complexdatacollective/Fresco/pull/27/commits/4060004f14b15c1fe24651053c902dffb29019dd)

There was a similar pattern in ProtocolUploader that I fixed here as well.

jthrilly (approved):

PR #37, opened 2023-12-04
Bug Fix: Proper handling of empty assets in `deleteProtocols`

mrkarimoff:
This PR fixes a bug in the `deleteProtocols` function. Previously, if there were no assets to delete, the function stopped execution in the first `try/catch` block with `return`. Now, with the fix, we check the case of no assets in the `deleteFilesFromUploadThing` function, and if true we stop the execution of code to delete assets from Uploadthing.


jthrilly (approved):
Excellent work!

PR #38, opened 2023-12-04
Bug Fix: Updating Asset model schema

mrkarimoff:
This PR fixes the `Invalid prisma.protocol.create() invocation` error by removing `@unique` from the `key` field of the Asset model.

Note: Prisma doesn't allow using `@id` and `@unique` on the same field
Issue here: https://github.com/prisma/prisma/issues/3236

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -28,7 +28,7 @@ model Protocol {
 }
 
 model Asset {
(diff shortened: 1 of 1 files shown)


jthrilly:
Great find @mrkarimoff!

jthrilly (approved):

PR #39, opened 2023-12-06
Feature: Floating feedback button

mrkarimoff:
This PR introduces the Feedback Button feature, allowing users to provide feedback or report bugs through a ClickUp form. 

The feature includes a floating button with a robot emoji, a tooltip with a helpful message, and a modal with an embedded ClickUp form for bug reporting.

The tooltip message disappears after 10 seconds and the robot emoji moves to the right corner of the screen.


components/FeedbackButton/FeedbackButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    return () => clearTimeout(timeoutId);
+  }, []);
+
+  return (
+    <>
+      <ClickUpFormModal open={open} setOpen={setOpen} />
+      <motion.div
+        initial={{ opacity: 1 }}
+        animate={{ x: '25px', rotate: -60 }}
```
buckhalt, line 25: In my opinion the rotated button here looks a bit odd. Is the idea just to move it out of the way after the tooltip disappears? I think it would be fine to animate it to the bottom corner without rotating.

components/FeedbackButton/ButtonTooltip.tsx
buckhalt, line 1: Any reason to not use the [shadcn tooltip ](https://ui.shadcn.com/docs/components/tooltip)here?

buckhalt (commented):
Cool feature! Left a couple of ideas below.

components/FeedbackButton/FeedbackButton.tsx
mrkarimoff: It looks creepy, right? 😅 I'll fix it.

components/FeedbackButton/ButtonTooltip.tsx
mrkarimoff: The reason I haven't used Shadcn tooltip is that I couldn't trigger it automatically, it works only on hover. I needed a tooltip that only appears once.

components/FeedbackButton/FeedbackButton.tsx
mrkarimoff: I put the feedback button in a top-level banner according to Joshua's suggestion.

buckhalt: The top-level banner looks awesome!

jthrilly (changes requested):
I don't really like how the form is looking inside of the modal window. The scrolling isn't a good user experience, and I don't like the duplication of the title.

Could you try putting it inside of a shadcn "sheet" instead?: https://ui.shadcn.com/docs/components/sheet

PR #42, opened 2023-12-18
Remove server code and references

buckhalt:
Fresco will not support Network Canvas Server pairing. This PR removes all Server code and references to it.


lib/interviewer/utils/exportProcess.js
jthrilly, line 1: Probably out of scope, but shouldn't this, and all export related files, be deleted rather than modified?

jthrilly (approved):
Looks good to me!

PR #48, opened 2024-01-11
Fix/promptid selectors

jthrilly:
This fixes bugs related to the lack of a promptId in certain contexts.

The issue was caused by faulty selectors that assumed promptIndex (note: not the same as prompt ID!) was being passed into the stage component as a prop. I removed this during refactoring as it would cause the stage to re-render when the prompt changed (so was incorrect).

The new implementation gets the promptId from the session, via the promptIndex. This is implemented within the `getPromptIndex` session selector.

Ultimately, the issue was manifesting in the `getPromptNodeModelData` selector, which is used to determine the metadata that nodes should have when they are created or added to a given stage/prompt combination. However it was also causing several other selectors to return incorrect data.


lib/interviewer/selectors/interface.js
```diff
@@ -117,9 +118,15 @@ export const makeGetStageNodeCount = () => {
...
+export const getPromptId = createSelector(
+  getPrompts,
+  getPromptIndex,
+  (prompts, promptIndex) => prompts[promptIndex].id,
+)
+
 export const getNetworkNodesForPrompt = createSelector(
   getNetworkNodesForType,
-  getPropPromptId,
+  getPromptId,
```
jthrilly, line 129: Replace the old selector with the new one - this selector was previously also failing without us realizing.

lib/interviewer/selectors/interface.js
```diff
@@ -133,7 +140,7 @@ export const makeNetworkNodesForPrompt = () => getNetworkNodesForPrompt;
 */
 
 export const getNetworkNodesForOtherPrompts = createSelector(
-  getNetworkNodesForType, getPropPromptId,
+  getNetworkNodesForType, getPromptId,
```
jthrilly, line 143: As above, update to use the session state selector rather than looking for prompt ID in props.

lib/interviewer/selectors/name-generator.js
```diff
@@ -23,13 +24,27 @@ const stageCardOptions = (_, props) => props.stage.cardOptions;
...
+  propStageId,
+  getPromptId,
+  (stageId, promptId) => {
+    return {
+      stageId,
+      promptId,
+    };
+  },
+)
+
```
jthrilly, line 37: A "helper" selector that just gets the stage and prompt ID and returns them in an object. This replaces the one I removed from the props selectors file, and puts it in a more appropriate location given that it now depends on state rather than props. I could also have put this in the interfaces selector file, and may do so if it turns out other interfaces use it.

lib/interviewer/selectors/name-generator.js
```diff
@@ -23,13 +24,27 @@ const stageCardOptions = (_, props) => props.stage.cardOptions;
...
-    ...ids,
-  }),
+  getIDs,
+  ({ type }, { stageId, promptId }) => {
+    return {
+      type,
+      stageId,
+      promptId,
+    };
+  },
```
jthrilly, line 47: This is the actual "fix" (although the real fix is in the new implementation of the getIDs selector, so as not to use props).

I also changed the style a little to remove the spread operator, which is something I want us to do more broadly. Spread operators can make it difficult to debug problems because it isn't clear which data they contain.

lib/interviewer/selectors/prop.js
```diff
@@ -22,7 +22,6 @@ export const getAdditionalAttributes = (stage, prompt) => {
 export const propStage = (_, props) => props?.stage ?? null;
 export const propPrompt = (_, props) => props?.prompt ?? null;
 export const propStageId = (_, props) => props?.stage?.id ?? null;
-export const getPropPromptId = (_, props) => props?.prompt?.id ?? null;
```
jthrilly, line 25: I removed this entirely because (1) it is broken now that prompt isn't consistently passed to props, and (2) I wanted to force myself to find other instances where this broken selector was used and fix them.

lib/interviewer/selectors/__tests__/interface.test.js
```diff
@@ -139,14 +139,6 @@ const mockState = {
...
 describe('interface selector', () => {
   describe('memoed selectors', () => {
-    it('makeGetIds()', () => {
-      const selected = Interface.makeGetIds();
-      expect(selected(mockState, mockProps)).toEqual({
-        stageId: mockStage.id,
-        promptId: mockPrompt.id,
-      });
-    });
-
```
jthrilly, line 149: Removed this test because I also deleted the underlying selector. See below.

buckhalt (dismissed):

lib/interviewer/selectors/name-generator.js
```diff
@@ -23,13 +24,27 @@ const stageCardOptions = (_, props) => props.stage.cardOptions;
...
 )
 
 export const makeGetPromptNodeModelData = () => getPromptModelData;
```
mrkarimoff, line 50: I didn't understand the purpose of this function because it's not being called in anywhere

mrkarimoff (commented):
Okay, this is a really interesting approach. I admit I haven't thought about getting the `promptId` via promptIndex from the session state

PR #53, opened 2024-01-18
Feature/suspenseful loading states

jthrilly:
This PR removes server side data fetching on dashboard routes, and replaces it with `Suspense` and `useSuspenseQuery` from tRPC.

Rationale is to experiment with reducing the time to first render. This allows the server rendered page.tsx to be static (was previously `force-dynamic`. 

Further work needs to be done to create appropriate loading shells for these routes.

Merge after https://github.com/complexdatacollective/Fresco/pull/44


jthrilly:
Closing this as no solution was easily available to the performance problems.

PR #54, opened 2024-01-18
Feature/interviewer scrolling

buckhalt:
This feature implements:
- Fixed full height for the interview, accounting for the banner
- Allowing scroll on Information Interfaces


jthrilly (approved):
Seems to work great! Not reviewed the implementation in depth, but we can fix later.

PR #56, opened 2024-01-19
Fix/hyperlist rendering

buckhalt:
The HyperList component is a virtualized list of data cards. The height of these data cards is calculated dynamically by using the react-dom server API `renderToString`.

Next 14 throws an error for importing this, despite the fact that the API is depreciated in React. This is because NextJS generates compiler aliases so that references to legacy browser code in react-dom/server throw errors.

This PR implements a workaround for this issue by bypassing the alias and maintaining the existing implementation.


jthrilly (approved):
Good PR (particularly the summary of the issue). Only thing I might have added would be the link where you found the magic import :)

PR #60, opened 2024-01-25
Update inctances of  `trackEvent`

mrkarimoff:
This PR updates every use case of `trackEvent` to adjust to the new error type of analytics package (`error` is a type of JS `Error` and also supports additional `metadata` property


jthrilly (approved):
Approving for now. I'm going to update trackEvent separately for better error handling when analytics fails.

PR #61, opened 2024-01-25
Feature/participant urls

buckhalt:
https://github.com/complexdatacollective/Fresco/pull/58 removed the "active" protocol concept. This also removed the ability to simply generate a participant URL without also having a protocol ID.

**This PR adds back in the ability to generate participation URLs for participants/protocols.**

Features:
- Ability to generate and export an anonymous recruitment URL (if enabled) for each protocol:
    - Within the anonymous recruitment section
    - From a column on the protocols table
- Ability to generate a CSV containing participation URLs 
    - From the protocols view: select protocol, select participants (defaults to all), export csv
    - From the participants view: select protocol, export csv
- Ability to generate a single participation url for each participant on the table. Can select the protocol from a dropdown on a modal


jthrilly (approved):
Implementation not reviewed in any depth, but the UX is great. Nice work on these changes.

PR #62, opened 2024-01-29
General bug fixes on Fresco interviewer

mrkarimoff:
This PR intends to provide the following fixes:
1. "BUG: `prompt` is undefined on stage 18" (the `prompt` is being undefined on the Sociogram when animating to the next stage)

The fix: 
- added conditional destructuring of the `prompt` and assigned an empty string to `layoutVariable` in case the `prompt` is undefined

2. Nodes not being rendered on Sociogram when `allowAutomaticLayout` is enabled

The fix: 
- removed `listener` from `initialize` function's useCallback dependency (not sure why it solved it)


lib/interviewer/selectors/canvas.js
```diff
@@ -13,7 +11,7 @@ import { get } from '../utils/lodash-replacements';
 import { getAllVariableUUIDsByEntity } from './protocol';
 import { getStageSubject } from './prop';
 
-const getLayout = (_, props) => get(props, 'prompt.layout.layoutVariable');
+const getLayout = (_, props) => get(props, 'prompt.layout.layoutVariable', '');
```
jthrilly, line 14: This should be null or undefined - not an empty string

PR #64, opened 2024-01-31
Fix/drag drop

buckhalt:
Drag/drop system was broken when it was refactored for Fresco. This PR fixes drag/drop by reverting components to Interviewer versions. It also ensures that the MultiNodeBucket rerenders after a node is dropped into a categorical bin so that already-dropped nodes do not still appear in the list.

Fixes drag/drop on:
- Name generator side panels
- Ordinal bins
- Categorical bins
- Deleting a node


jthrilly (approved):
Looks good to me. Like the solution with the component display name - I've never seen that before!

PR #73, opened 2024-02-01
fix/dyad census and tie strength census

buckhalt:
When there are no steps and multiple prompts on Dyad Census and Tie Strength Census interfaces, navigation back/forward should ignore multiple prompts and force next/prev _stage_. 

With the refactor of navigation and implementation of `useNavigationHelpers`, this functionality was removed. The default navigation prompted navigation to the next/prev _prompt_ instead. This caused an issue where next/prev navigation buttons appeared to need to be clicked multiple times to advance past the introduction, once for each prompt.

This PR fixes navigation for dyad census and tie strength census by forcing navigation to next/prev _stage_ if there are no steps. This is implemented using an optional `options` prop in the `moveBackward` and `moveForward` methods in `useNavigationHelpers`.

Runtime errors on these interfaces are also fixed here.

### DIFF
--- a/lib/interviewer/containers/Interfaces/DyadCensus/DyadCensus.js
+++ b/lib/interviewer/containers/Interfaces/DyadCensus/DyadCensus.js
@@ -89,7 +89,7 @@ const DyadCensus = ({
     if (isIntroduction) {
       // If there are no steps, clicking next should advance the stage
       if (stepsState.totalSteps === 0) {
-        navigationActions.moveForward();
+        navigationActions.moveForward({ forceChangeStage: true });
         return;
       }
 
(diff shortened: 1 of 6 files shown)


jthrilly (approved):
Really well done for figuring this out.

I don't _love_ the idea of polluting the function signature for `moveForward`/`moveBackward` with options, but I see why it has to be done. 

There are two things to think about:
1. Is `options` and then `forceStageChange` right right structure and naming convention? Here I would ask if we are likely to add more "options" in the future (my thinking is probably not), and then if not could I make it a single function parameter perhaps just simply called `force`. We would get some IDE autocomplete on then when using it, which would work nicely.
2. The other approach, which could actually be combined with the approach above, would be to have the hook export alternative versions of the navigation methods that don't perform the prompt check. You could call these `forceMoveBackward` and `forceMoveForward`. They could be implemented internally using a variable as in (1), but this implementation detail could be hidden from the consumer. The obvious downside to this is more verbiosity.

Just posting these considerations as stuff to think about if you didn't already! Going to merge this in as-is.

PR #77, opened 2024-02-13
Fix/narrative interface

buckhalt:
This PR fixes several issues on the Narrative Interface:
- Missing attribute labels in the preset switcher
- Missing groups section in the preset switcher
- Edges not displaying on interface
- Convex hulls not displaying on interface
- Preset Switcher overflowing screen when all options are open
- Node attributes radio button distorted when there is a long variable name

### DIFF
--- a/lib/interviewer/containers/Canvas/PresetSwitcherKey.js
+++ b/lib/interviewer/containers/Canvas/PresetSwitcherKey.js
@@ -15,6 +15,7 @@ import {
 } from '../../selectors/network';
 import { get } from '../../utils/lodash-replacements';
 import { createPortal } from 'react-dom';
+import { getCurrentStage } from '../../selectors/session';
 
 class PresetSwitcherKey extends Component {
   constructor() {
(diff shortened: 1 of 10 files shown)


lib/interviewer/containers/Interfaces/Narrative.js
```diff
@@ -225,9 +230,6 @@ function mapStateToProps(state) {
   };
 }
 
-const mapDispatchToProps = () => ({
-});
+const mapDispatchToProps = () => ({});
```
jthrilly, line 233: I _think_ you can remove this and omit the second parameter below

buckhalt: resolved [3301d7b](https://github.com/complexdatacollective/Fresco/pull/77/commits/3301d7b56f5109648b4414804bc19c4c953307dc)

lib/interviewer/containers/Canvas/PresetSwitcherKey.js
```diff
@@ -125,13 +129,18 @@ const makeMapStateToProps = () => {
 
   const mapStateToProps = (state, props) => {
     const highlightLabels = get(props, 'preset.highlight', []).map((variable) =>
-      getNodeAttributeLabel(state, { variableId: variable, ...props }),
+      getNodeAttributeLabel(state, {
+        stage: getCurrentStage(state),
```
jthrilly, line 133: given that this is happening inside a `map` it would be good to just calculate the current stage once, and then use that variable inside of the map. You can just to `const stage = getCurrentState(state)` in side of `mapStateToProps`, and then reuse it.

buckhalt: resolved [a44327c](https://github.com/complexdatacollective/Fresco/pull/77/commits/a44327c33624fcce25e17666f987c8989ff2f1fa)

PR #78, opened 2024-02-15
Fix: beforeNextFunction null on quick consecutive navigation

buckhalt:
**Bug**:
- Quick consecutive navigation allows bypass of skip logic
- This is because calling moveForward or moveBackward before the stage has fully rendered means that the stage has not yet called registerBeforeNext. So beforeNextFunction is null, and the stage can be skipped.
- This happens because Stage has entry and exit animation that takes time to complete.

**Fix**:
- Disable navigation while animating the Stage transition.


jthrilly (changes requested):
I think this is a cool approach, but the implementation could be better.

Some problems with this:
- We've polluted the disabled state of the navigation buttons. Previously we were checking if we can navigate in the direction of the button, but now we are checking for an additional property. Isn't the additional property (in this case the fact we are animating) an _aspect_ of if we can navigate?
- We've got two separate state items that are holding the same information (in `<Stage>` and `<ProtocolScreen>`. To make things worse, we are passing the setState function of one into the other, and then synchronising them using a `useEffect`. Bad code smell! :) The need for a prop such as `onIsAnimatingChange` should be a clue that something has gone awry. There should be a single source of this state.

So with that said, better design for this is to put the animating state inside of `useNavigationHelpers`. This solves all the issues above:
- You can include the current animation state in the calculation for `canMoveForward` | `canMoveBackward`.
- You don't need two separate `useState`s that you then have to sync. You can use the state inside of `Navigation`, and you can set it inside of `Stage`. The `ProtocolScreen` component doesn't need to be involved.

It might be better to make the naming more general purpose - maybe call the setter something like `setForceNavigationDisabled(value: boolean)`. I don't see a need right now to expose the raw state value, since for this use case it can be included in the value of `canMoveBackward` etc.

buckhalt:
Refactored as requested:
- Moved animating state inside of `useNavigationHelpers`
- Renamed state to be more general purpose - `forceNavigationDisabled` instead of `isAnimating`
- used `jotai` for state management
- use exposed `setForceNavigationDisabled` in `Stage` to set it
- included `forceNavigationDisabled` in the calculation for `canMoveForward` | `canMoveBackward` within `useNavigationHelpers` so that raw state value does not need to be exposed.

PR #82, opened 2024-02-20
Finish Interview Improvements

buckhalt:
This PR implements several improvements to the Finish Interview functionality:

**Confirm dialog UI/UX Improvements**
- Refactored to use Interview dialog system
- Loading spinner while finish is processing added

**Event tracking for InterviewCompleted**
- `nodeCount` and `edgeCount` metadata added

**Option to limit participants to one interview per protocol**
- `limitInterviews` setting added to `appSettings`. Toggleable from app setup and settings page on the dashboard
- Limiting achieved using cookies. Cookie is set when interview is completed, with the `protocolId` as the key. There are checks for the cookie in the `/interview` and `/onboard` routes. If cookie is present and `limitInterviews` is true, user is redirected to the thank you for participating page.


lib/interviewer/containers/Interfaces/FinishSession.js
```diff
@@ -1,23 +1,70 @@
-import { useEffect, useState } from 'react';
+import { useState, useEffect, useCallback } from 'react';
 import { useDispatch } from 'react-redux';
 import Button from '~/lib/ui/components/Button';
-import FinishInterviewModal from '~/app/(interview)/interview/_components/FinishInterviewModal';
+import { actionCreators as dialogActions } from '../../ducks/modules/dialogs';
```
mrkarimoff, line 4: Would be nice to fix the import path here.

mrkarimoff (changes requested):
Really great work, man 👏 .  I asked for a minor change above. If you think this isn't necessary, feel free to mark it as resolved

lib/interviewer/containers/Interfaces/FinishSession.js
buckhalt: resolved [7e84cd5](https://github.com/complexdatacollective/Fresco/pull/82/commits/7e84cd559e6db0a8cf986b0cbd2f758c79a3cca7)

app/(interview)/onboard/[protocolId]/route.ts
```diff
@@ -13,6 +15,15 @@ const handler = async (
...
   }
 
+  const appSettings = await api.appSettings.get.query();
+
+  // if limitInterviews is enabled
+  // Check cookies for interview already completed for this user for this protocol
+  // and redirect to finished page
+  if (appSettings?.limitInterviews && cookies().get(protocolId)) {
+    redirect('/interview/finished');
+  }
```
Comment (jthrilly) [inline app/(interview)/onboard/[protocolId]/route.ts:25]: In future, think about making this a single function in a separate file, that the route handler and the page can both import. That way we can update the behaviour later in a simpler way.

jthrilly (approved):

PR #90, opened 2024-02-29
Fix: Interview Export Errors

buckhalt:
Exporting interview sessions was not working properly on Safari and Firefox. This was due to the way the useDownload hook was implemented and how it creates an "a" element to trigger the download.

On some browsers, there are increased security restrictions around programmatically initiated downloads. The way to work with this is to create a blob and use its url as the href for the "a" element. Other uses of useDownload in this project create the blob before calling the download method, and pass it the blob's url. ExportInterviewsDialog was not implemented in the same way.

Therefore, this fix implements using blob and its url in ExportInterviewsDialog which fixes the issue. Future refactoring could be done to move this functionality all within the useDownload hook.

### DIFF
--- a/components/ErrorReportNotifier.tsx
+++ b/components/ErrorReportNotifier.tsx
@@ -67,11 +67,6 @@ export default function ErrorReportNotifier({ error }: { error: Error }) {
     if (initialized.current) return;
     setState('loading');
 
-    // limit size of error.stack
-    if (error.stack && error.stack.length > 500) {
-      error.stack = error.stack.substring(0, 500);
-    }
-
     trackEvent({
       type: 'Error',
       name: error.name,
--- a/components/ErrorReportNotifier.tsx
+++ b/components/ErrorReportNotifier.tsx
@@ -67,6 +67,11 @@ export default function ErrorReportNotifier({ error }: { error: Error }) {
     if (initialized.current) return;
     setState('loading');
 
+    // limit size of error.stack
+    if (error.stack && error.stack.length > 500) {
+      error.stack = error.stack.substring(0, 500);
+    }
+
     trackEvent({
       type: 'Error',
       name: error.name,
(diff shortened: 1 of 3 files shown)


components/ErrorReportNotifier.tsx
```diff
@@ -67,6 +67,11 @@ export default function ErrorReportNotifier({ error }: { error: Error }) {
     if (initialized.current) return;
     setState('loading');
 
+    // limit size of error.stack
+    if (error.stack && error.stack.length > 500) {
+      error.stack = error.stack.substring(0, 500);
+    }
+
```
jthrilly, line 74: I'm not so sure about this. The spec says 64kb is allowed as the payload size for in flight (as in, not yet delivered) keepalive fetch requests (https://github.com/whatwg/fetch/issues/679). You're truncating the stack to 500 characters (I think...it isn't clear to me how length interacts with strings that can contain multi-byte unicode characters) which is way less than 64kb.

So I think what is actually happening here is that the error with the anchor tag construction was calling the error reporting endpoint multiple times in quick succession. Based on the stack trace in the screenshots you shared, that could be because of re-rendering behaviour in the development environment, or because of an infinite loop of some kind. Either way, this was filling up the buffer, and triggering this second "error".

So there are a couple of things:
1. I don't think arbitrarily truncating the stack is a good idea. We would miss out on data.
2. The fact that there is a difference between dev and production is likely due to differences in error handling behaviour in nextJS. In production, the stack is actually replaced with a much smaller string that just contains a pointer to a hash in the log (see: https://nextjs.org/docs/app/building-your-application/routing/error-handling#securing-sensitive-error-information).
3. The correct way to handle the in flight payload size issue is to handle the fetch request failing gracefully. This would be in `createRouteHandler` in the analytics package, but the code there looks like it should handle it already.

My suggestion for now is just to fix the anchor tag construction and ignore the error handling stuff for now.

jthrilly (changes requested):

components/ErrorReportNotifier.tsx
buckhalt: Sounds good to me. Reverted error handling stuff for now - 223f181

PR #93, opened 2024-03-19
fix: assetId not unique

buckhalt:
Some protocols were failing to import because of a duplicate protocol issue. This was due to the way we use the assetManifest key as the assetId. We cannot actually assume that the assetManifest key is unique given that we allow protocols that are copies of one another if they have made changes.

This PR makes the assetId truly unique by appending a cuid to the key.

### DIFF
--- a/utils/protocolImport.tsx
+++ b/utils/protocolImport.tsx
@@ -1,5 +1,6 @@
 import type { Protocol } from '@codaco/shared-consts';
 import type Zip from 'jszip';
+import { createId } from '@paralleldrive/cuid2';
 
 // Fetch protocol.json as a parsed object from the protocol zip.
 export const getProtocolJson = async (protocolZip: Zip) => {
@@ -63,7 +64,7 @@ export const getProtocolAssets = async (
       }
 
       files.push({
-        assetId: key,
+        assetId: `${key}_${createId()}`, // We cannot assume key is unique across protocols. Add a unique suffix.
         name: asset.source,
         type: asset.type,
         file: new File([file], asset.source), // Convert Blob to File with filename


jthrilly:
This is an interesting one. I think this hinges on one question: are "forked" assets, which are identical to one another in terms of content, and which share the same UUID, different assets? The way you fixed this assumes that they are, but my view is that they aren't. The asset UUID is generated by Architect when an asset is created. It can't be changed. and it has pretty strong likelihood of global uniqueness. As always, a UUID intends for a record to be unique _within a system_ first and foremost. and it clearly passes that test.

From that perspective, the fact that duplicated protocols have assets that share a UUID is actually a feature, not a bug. So given that, the real issue here is that we are failing on insertion when an asset with a UID already exists, rather than simply linking to it and saving ourself from creating a new db record and S3 bucket item.

There are two obvious problems with doing this, though: 

1. What is there is a natural collision? This seems extremely unlikely unless someone manually authored a protocol. 
2. What happens when deleting protocols - how do we know if we can safely delete any media assets? This would require checking to see if an asset is linked to any other Protocol records at delete time.

So either way we implement this we have a drawback. With your approach we are being space inefficient but we can track asset deletion. With my approach we are (I believe) maintaining the correct behaviour of the system, but we add a lot of complexity to the upload system (which has to work very differently if we can't just assume that assets can be uploaded right after validation completes) and the delete protocol functionality.

Aside from that, there's one reason I don't like your fix specifically though: we are taking a UUID that we have already generated, but now _know_ not to be unique, and we are just adding on some more uniqueness to the end. Why bother using it at all? Why not have our own unique identifier for _this_ system, and store the current UUID in a different column (if we arent already) with more relaxed constraints?

buckhalt:
Refactored the protocol upload and deletion functionality to allow protocols with assets that already exist to be uploaded. 

This change involves:
- Creating a many-to-many relationship between Protocol and Asset
- Protocol import: checking if assets already exist, and if so, connecting them to the protocol. If asset is not yet in the database, create a database record and S3 bucket item.
- Protocol deletion: select assets that are ONLY associated with the protocols to be deleted before deleting assets. If an asset is used by a protocol not being deleted, leave it in the database.

mrkarimoff (approved):
That looks great! Took me a little longer to understand, probably because I haven't coded in Fresco for a while 😄. I think we are handling the deleting process much efficient now. Anyway, I didn't spot any issue here, good work Caden

hooks/useProtocolImport.tsx
```diff
@@ -160,9 +162,47 @@ export const useProtocolImport = () => {
...
+      const assetsWithCombinedMetadata: z.infer<typeof assetInsertSchema> = [];
+
+      let newAssetsWithCombinedMetadata: z.infer<typeof assetInsertSchema> = [];
+
+      // Check if the assets are already in the database.
+      // If yes, add them to assetsWithCombinedMetadata to be connected to the protocol.
+      // If not, add them to newAssets to be uploaded.
 
       if (assets.length > 0) {
+        const assetPromises = assets.map(async (asset) => {
```
jthrilly, line 177: I _kind of_ get why you did it this way, but I also think we should make this be more efficient. 

Right now you're creating a separate query for every asset in the protocol. You are then immediately invoking all of these queries, and waiting for them all to settle. There are a few problems with that:

1. All the queries are competing with each other, since they are all immediately invoked. This means all network requests will be sent at the same time, regardless of what else the client or the server are doing. This will lead to performance issues eventually, and is almost always not what you intended. If you wanted to do it this way, a queue with some managed concurrency would be better.
2. The way you've structured this with a `promise.all` means that any individual failure with any of these (many, many!) queries will abort the whole import. Its basically impossible to do any error handling.
3. One invokation of the tRPC endpoint per-asset is going to be very resource intensive on the database server, too. It will have to read all rows in the table for each query, and subsequent queries won't be able to benefit from any information from previous ones.

Here's what I suggest:
1. Refactor the router to accept an array of IDs to check, and return something useful like the IDs that don't already exist. If you want it to be a more general-purpose endpoint, you could return either an object with keys corresponding to each array item and a boolean value, OR something like `{ found: [123,456], notFound: [789] }`.
2. Call this endpoint inside a try/catch block with all protocol assets in one go, so that you can handle the error case that this check failed.
3. Tidy up the code following this - I don't like the look of L:195!

buckhalt: refactored using this approach here [a3be68e](https://github.com/complexdatacollective/Fresco/pull/93/commits/a3be68e49cdca722f050a9a157639cfdd0f4cf87)

PR #101, opened 2024-04-08
Fix/dashboard bundle size

mrkarimoff:
This PR aims to reduce the bundle size on `/dashboard` page through these changes:

- Adds server-side `getActivities` query to fetch initial activity data for the first page on `/dashboard`
- Passes the initial data to `ActivityFeed` via props to use it on the client-side query
- Moves import of `@faker-js/faker` in the specific utility functions
- Bumps the version to `-beta.3`

The bundle size gets reduced from `1.26 MB` to `261 kB` ✅


jthrilly:
Closing as I have addressed this in my docker branch in a cleaner way.

PR #113, opened 2024-04-16
Feature/delete uploadthing interviews

mrkarimoff:
This PR introduces the following changes/features

- Updates UploadThing SDK and @uploadthing/react to the latest
- Implements automatically deleting the interview session zip file from the UploadThing bucket after download (with 3s delay)


app/(dashboard)/dashboard/interviews/_components/ExportInterviewsDialog.tsx
```diff
@@ -84,6 +85,11 @@ export const ExportInterviewsDialog = ({
       download(url, result.data.name);
       // clean up the URL object
       URL.revokeObjectURL(url);
+
+      // Delete the zip file from UploadThing after 3 seconds
+      setTimeout(async () => {
```
buckhalt, line 90: What happens if the download takes longer than 3 seconds? Will the asset be deleted before it finishes downloading?

Can we somehow programmatically detect when it has been downloaded, and call the delete function then instead?

buckhalt (changes requested):
Thanks for this Mirfayz! See my comments for a requested change.

app/(dashboard)/dashboard/interviews/_components/ExportInterviewsDialog.tsx
mrkarimoff: Okay, I did some research on it. Turns out, it doesn't matter if the zip file gets deleted before a user downloads it: **Once you have a Blob object, it means you have the file data locally in your browser's memory. A Blob object represents a chunk of bytes that holds data of a file.** , so I added a check for the UploadThing response to make sure it doesn't fail.

app/(dashboard)/dashboard/interviews/_actions/deleteZipFromUploadThing.ts
```diff
@@ -0,0 +1,14 @@
+'use server';
+
+// We need to delete the exported data from UploadThing after the user has downloaded it.
+// This is to ensure that we are not storing any sensitive data on UploadThing for longer than necessary.
+
+import { utapi } from '~/app/api/uploadthing/core';
+
+export const deleteZipFromUploadThing = async (key: string) => {
+  const deleteResponse = await utapi.deleteFiles(key);
```
jthrilly, line 9: There's no security on this action, as it doesn't require auth. Doesn't this mean that a malicious actor could spam the endpoint and try to delete data?

Its obviously unlikely, because file keys are UUIDs, but it still seems important...

app/(dashboard)/dashboard/interviews/_components/ExportInterviewsDialog.tsx
jthrilly: Where did you remove the timeout? I don't see a commit for it.

mrkarimoff: > Where did you remove the timeout? I don't see a commit for it.

Here's the commit : 
https://github.com/complexdatacollective/Fresco/pull/113/commits/b62da1a4ba5b8d47d6ece161e15b2ed36638cb98

app/(dashboard)/dashboard/interviews/_actions/deleteZipFromUploadThing.ts
mrkarimoff: Good point, just added auth check here: https://github.com/complexdatacollective/Fresco/pull/113/commits/e0929fd6539203839879bb54d2e738dfcdb42a8b

mrkarimoff: Sorry about this, here's the proper auth check: https://github.com/complexdatacollective/Fresco/pull/113/commits/db1b8d48cde89cb2ad1dd68a7e07eb037dd159da

PR #118, opened 2024-04-29
Experiment/remove trpc

jthrilly:
This PR demonstrates a radical rewrite of Fresco to use approaches promoted in the nextjs documentation:

- all data fetching done in RSCs and passed into client components where required
- top level data fetching from pages removed in favour of fetching inside (server) components
- suspense boundaries wrapping these components to allow their content to be streamed in
- use of server actions for data mutation
- data functions are split into queries and actions, with files corresponding to the different entity types being manipulated
- implementation of page level auth requirements, as well as API level auth helpers

it also implements 
- complete removal of client redirect wrapper, and replacement with a new, much cleaner, redirect system
- lucia auth v3

known issues:

- [ ] Interview navigation should push to browser history so back and forward buttons work
- [x] hiding the feedback banner in the interview redirects to the login page, not the current interview
- [x] redirection to the signin page doesn't include the URL to redirect back to (for example callbackURL=/dashboard/settings). Needs reimplementing.
- [ ] There is a visual glitch when loading a name generator. it looks like the nodes suddenly move upwards after animating in.
- [x] Finish interviews not working
- [x] Need to reimplement LimitInterviews and AnonymousRecruitment switches in onboarding step
- [x] Resetting app settings redirects to the login page, not the setup page
- [x] Generate participation urls select protocol not getting options
- [ ] We need a plan for data migration - either we go back to lucia auth v2, or we figure out how to run migrations in production. Interestingly, drizzle can do this seemingly without any issues: https://youtube.com/watch?v=dHTGEQnogPw. It keeps track of which migrations have been applied in a special migrations table in the DB. This is so obvious that I can't believe prisma doesn't also do it. Have we missed something?
- [x] Delete participants modal does not close after confirm
- [ ] Activity table pagination, type, filter, etc is not re-rendering the table without a hard refresh


utils/auth.ts
```diff
@@ -1,33 +1,98 @@
...
-  return authRequest.validate();
-});
+export async function requirePageAuth(
+  { redirectPath = null } = {} as RequireAuthOptions,
+) {
+  const { session } = await getServerSession();
+
+  if (!session) {
+    if (!redirectPath) {
+      redirect('/signin');
```
buckhalt, line 82: why is this redirect being called in an if statement when it's going to be called regardless of the condition right after?

jthrilly:
Merging this now, as it has simply become too big!

PR #122, opened 2024-04-30
potential workaround for uploadthing file deletion crashing

jthrilly:
This PR changes the way temp file deletion works during the export process.

- It calls the deleteZipFromUploadThing method in the `finally` block, which both prevents deletion from cancelling the export, and allows us to potentially clean up even if the export fails.
- It adds a user toast with messaging in the scenario that the deletion files, so that the user can manually delete any sensitive data.


jthrilly:
Superseded by https://github.com/complexdatacollective/Fresco/pull/124

PR #131, opened 2024-05-13
Beta 5

jthrilly:
Outstanding tasks from RSC refactor:

- [x] **Remove history push from navigation buttons**
- [x] There is a visual glitch when loading a name generator. it looks like the nodes suddenly move upwards after animating in.
- [x] We need a plan for data migration - either we go back to lucia auth v2, or we figure out how to run migrations in production. Interestingly, drizzle can do this seemingly without any issues: https://youtube.com/watch?v=dHTGEQnogPw. It keeps track of which migrations have been applied in a special migrations table in the DB. This is so obvious that I can't believe prisma doesn't also do it. Have we missed something?
- [x] ~Activity table pagination, type, filter, etc is not re-rendering the table without a hard refresh~ (separate PR now)
- [x] Reimplement error messaging for incorrect username & password
- [x] Delete participant confirmation dialog doesn't close after clicking continue
- [x] Importing multiple protocols at the same time causes all import toasts to re-render any time any of them updates.
- [x] Limit interviews not working
- [x] Exporting interview with 0% progress fails


jthrilly:
Via @buckhalt:

- [ ] changes to the analytics package are causing errors in prod due to not having access to window.location. this is the cause of one bug on file export, creating an interview, and finishing an interview
- [ ] exporting data was still failing due to uploadthing bug which is fixed by updating that package
- [ ] there’s an issue with the getBaseUrl funciton and its use of VERCEL_URL. we need to rethink how we’re getting the url.
- [ ] there may still be an issue with gateway timeout on analytics reporting. this could just be related to the issues with the analytics package update though.

PR #132, opened 2024-05-14
Reimplement login form error handling

buckhalt:
includes action error component, input error messaging and uses next-safe-action for safeLogin action

### DIFF
--- a/app/(blobs)/(setup)/_components/SignInForm.tsx
+++ b/app/(blobs)/(setup)/_components/SignInForm.tsx
@@ -1,0 +1,87 @@
+'use client';
+
+import { useState } from 'react';
+import { safeLogin } from '~/actions/auth';
+import ActionError from '~/components/ActionError';
+import { Input } from '~/components/ui/Input';
+import SubmitButton from '~/components/ui/SubmitButton';
+import useZodForm from '~/hooks/useZodForm';
+import { loginSchema } from '~/schemas/auth';
+
+type ResponseError = {
+  title: string;
+  description: string;
+};
+
+export const SignInForm = () => {
+  const {
+    register,
+    handleSubmit,
+    formState: { errors },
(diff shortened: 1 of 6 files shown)


app/(blobs)/(setup)/_components/SignInForm.tsx
```diff
@@ -0,0 +1,83 @@
+'use client';
+
+import { Loader2 } from 'lucide-react';
+import { useRouter } from 'next/navigation';
+import { login } from '~/actions/auth';
+import { Button } from '~/components/ui/Button';
+import { Input } from '~/components/ui/Input';
+import { useToast } from '~/components/ui/use-toast';
+import useZodForm from '~/hooks/useZodForm';
+import { loginFormSchema } from './schemas';
```
jthrilly, line 10: This should use the existing schema in `~/schemas/auth`.

app/(blobs)/(setup)/_components/schemas.ts
```diff
@@ -22,6 +22,11 @@ export const userCreateFormSchema = z.object({
   ),
 });
 
+export const loginFormSchema = z.object({
+  username: z.string().min(1, { message: 'Username cannot be empty' }),
+  password: z.string().min(1, { message: 'Password cannot be empty' }),
+});
+
```
jthrilly, line 29: shouldn't be here. already in `~/schemas/auth`.

app/(blobs)/(setup)/_components/schemas.ts
jthrilly, line 1: this filename doesn't really make sense given the content. if the schema is moved to where it should be, its just a single set of classes. These should either be with the component that first uses them, or this file should be renamed to `containerClasses`.

jthrilly (changes requested):

PR #133, opened 2024-05-15
Feature/docker

jthrilly:
This PR will introduce a dockerised version of the app, including deployment of db + app together via docker compose. 

It also implements automatic database migration.

**Note:** This requires documentation to be updated to cover changed environment variable names!


jthrilly:
Currently working on handling the scenario where existing users don't have a migrations table.

jthrilly:
Last task is to get a docker image building in github actions.

jthrilly:
Merging this now, as it contains many other improvements.

Key outstanding issue is that upload thing fails with protocol import. I can't investigate this further without having a host to deploy this on. Waiting on FSM IT for this.

PR #134, opened 2024-05-16
Fix activity feed

buckhalt:
Activity feed table was not updating with changes to the search params.

This fixes that issue by calling `router.refresh()` after updating the params and including the hashed params object in the cache key.


jthrilly:
Superseded by #135

PR #135, opened 2024-05-21
reimplement activity feed table as client component

jthrilly:
This PR implements the activity feed table as a client component with tanstack query and a single API route handler, that uses the existing query.

### DIFF
--- a/app/dashboard/_components/ActivityFeed/ActivityFeed.tsx
+++ b/app/dashboard/_components/ActivityFeed/ActivityFeed.tsx
@@ -1,18 +1,28 @@
+'use client';
+
+import type { Events } from '@prisma/client';
+import { useQuery } from '@tanstack/react-query';
+import { useSearchParams } from 'next/navigation';
 import { DataTableSkeleton } from '~/components/data-table/data-table-skeleton';
 import ActivityFeedTable from './ActivityFeedTable';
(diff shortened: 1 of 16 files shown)


app/dashboard/_components/ActivityFeed/ActivityFeed.tsx
```diff
@@ -1,18 +1,28 @@
...
-import { getActivities } from '~/queries/activityFeed';
 
-export default function ActivityFeed() {
-  const searchParams = searchParamsCache.all();
-  const activitiesPromise = getActivities(searchParams);
+export const ActivityFeed = () => {
+  const params = useSearchParams();
+
+  const { isPending, data } = useQuery({
+    queryKey: ['activityFeed', params.toString()],
```
buckhalt, line 13: Should the params be hashed for the queryKey?

jthrilly: Not in this case. Unlike with the next cache, there aren't any concerns about key length, because tanstack query hashes internally itself: https://tanstack.com/query/latest/docs/framework/react/guides/query-keys#query-keys-are-hashed-deterministically

The only requirement is that the key be serialisable, which SearchParams isn't by default, but is when converted to a string.

PR #136, opened 2024-05-23
reimplement recruitment test section

buckhalt:
Reimplements recruitment test section. Implemented using a server component to create promises for data queries and passing those promises to a client component which uses the use hook to read the values.


mrkarimoff (approved):
Tested all the features, and the code looks great to me 🎉

PR #137, opened 2024-05-24
fix: node animation in NameGenerator

mrkarimoff:
This PR fixes node animation with the following changes:
- Updates framer-motion to the latest (`11.2.6`)
- Removes the `layout` attribute in `NodeTransition`  component

(FYI: I tried everything and the only solutions that worked were either setting `y: '20%'` rather than `y: '3rem'` or removing the `layout` attribute)

Open to any suggestions!


jthrilly:
Removing the layout element will prevent nodes from animating position when nodes enter or leave the container. That's not a viable solution.

PR #138, opened 2024-05-28
Feature/dnd improvements

jthrilly:
This PR rewrites the drag and drop functionality entirely, using the [HTML Drag and Drop](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API) specification.

This should give us dramatically improved performance, due to removing the redux store. It won't support drop obstacles in the same way, but I see this as a reasonable tradeoff.

Current status:
- [X] Proof of concept using name generator interface
- [X] Implementation of zuzstand store for state
- [X] Custom hook for droppable, and HOC for draggable
- [X] Basic 'accepts', 'active', 'hover' 'valid' states.
- [ ] Refactoring 'accepts' for more complex scenarios, where a function needs to be called.
- [ ] Reimplement other interfaces
- [ ] Reimplement free dragging on the sociogram


jthrilly:
Closing in order to implement in Studio

PR #140, opened 2024-05-28
Fix participant label and identifier validation

buckhalt:
**This PR fixes a variety of issues related to participant label and identifier validation and type.** It assumes:
- `label` is an optional property
- Absence of `label` should be `undefined` in the database, not an empty string
- `caseProperty`, which is assigned the `label`, should therefore be optional in data export

**Changelog:**
- Makes `label` undefined instead of `"" `when no label is provided in add participant form and add via CSV file. Fixes bug where participant name in activity table was "".
- Improves validation schemas:
    - prevent whitespace as `label` and `identifier`
    - prevent leading and trailing whitespace on identifier, fixing URL issues.
- Replaces simplistic validation in CSV import with roboust validation schemas for consistency across methods for adding participants
- Fixes import CSV bug where importing a CSV fails when there is a `label` value that is blank, even if those rows have `identifier` values.
- Fixes bug where editing a participant gave a unique identifier validation error due to the editing participant already being in the db
- Fixes data export bug when no `identifier` is provided by making `caseProperty` optional. Necessary fix here because this was only working when identifier was an empty string instead of undefined.
- Adds tests to cover adding participants


lib/network-exporters/utils/general.js
```diff
@@ -52,7 +52,12 @@ export const getFileExtension = (formatterType) => {
  * Generate a filename prefix based on the session in the format:
  * `{caseId}_{sessionId}` 
  */
-export const getFilePrefix = (session) => `${sanitizeFilename(session.sessionVariables[caseProperty])}_${session.sessionVariables[sessionProperty]}`;
+export const getFilePrefix = (session) => {
+  if (!session.sessionVariables[caseProperty]) {
+    return `${sanitizeFilename(session.sessionVariables[sessionProperty])}`;
+  }
+  return `${sanitizeFilename(session.sessionVariables[caseProperty])}_${session.sessionVariables[sessionProperty]}`;
+}
```
jthrilly, line 60: Now that `caseProperty` can potentially be null, this is perhaps best implemented as nullish coalescence:

```
const caseIdentifier = session.sessionVariables[caseProperty] ?? session.sessionVariables[sessionProperty];

${sanitizeFilename(caseIdentifier)}
```

Up to you if you think this is more readable, but personally I think it is.

schemas/participant.ts
```diff
@@ -3,18 +3,49 @@ import { z } from 'zod';
 export const participantIdentifierSchema = z
   .string()
   .min(1, { message: 'Identifier cannot be empty' })
-  .max(255, { message: 'Identifier too long. Maxiumum of 255 characters.' });
+  .max(255, { message: 'Identifier too long. Maximum of 255 characters.' })
+  // Ensure that the identifier is not just whitespace or have leading/trailing spaces
+  .refine((identifier) => identifier.trim() === identifier, {
+    message: 'Identifier cannot have leading or trailing spaces',
```
jthrilly, line 9: This `refine()` checks that the identifier doesn't only consist of spaces, but the message implies that it also detects leading and trailing spaces. The message should say "Identifier requires one or more non-whitespace characters" (or similar).

Its fine to trim leading and trailing whitespace without telling the user. To do this, you need to add `.trim()` to the end of the expression.

schemas/participant.ts
```diff
@@ -3,18 +3,49 @@ import { z } from 'zod';
...
-export const participantLabelSchema = z.string().optional();
+const participantIdentifierOptionalSchema = z
+  .string()
+  .max(255, {
+    message: 'Identifier too long. Maximum of 255 characters.',
+  })
+  .optional()
+  .refine(
+    (identifier) =>
+      identifier === undefined || identifier.trim() === identifier,
```
jthrilly, line 20: I don't think the check for `identifier === undefined` is needed here, but if it _is_ (because of `optional()`) it might be better to check if identifier is defined (returning early if it isnt) and then implement a purer version of the "check for only whitespace" code. I think this will make it more readable.

You should consider making the whitespace check a utility function so it can be reused. It could follow the pattern of the other zod utilities.

I also think the message here is misleading as per my comments above. It should be updated the same way, with `.trim()` added at the end.

jthrilly (changes requested):
Looks great! Just a few nitpicks to resolve, then we can merge.

schemas/participant.ts
buckhalt: resolved [7141e0c](https://github.com/complexdatacollective/Fresco/pull/140/commits/7141e0c494fae69109e23f0b2b45482981dbae36)

schemas/participant.ts
buckhalt: resolved [7141e0c](https://github.com/complexdatacollective/Fresco/pull/140/commits/7141e0c494fae69109e23f0b2b45482981dbae36)

schemas/participant.ts
```diff
@@ -1,40 +1,40 @@
...
 export const participantIdentifierSchema = z
   .string()
+  .trim()
   .min(1, { message: 'Identifier cannot be empty' })
   .max(255, { message: 'Identifier too long. Maximum of 255 characters.' })
-  // Ensure that the identifier is not just whitespace or have leading/trailing spaces
-  .refine((identifier) => identifier.trim() === identifier, {
-    message: 'Identifier cannot have leading or trailing spaces',
+  .refine(hasNonWhitespaceCharacters, {
+    message: 'Identifier requires one or more non-whitespace characters.',
```
jthrilly, line 13: Won't calling `trim()` first mean that a string consisting of a space will have that space removed, so it will then fail with "identifier cannot be empty"? I think you want to call trim at the end.

Given we have vitest running now, why not write some tests for this parser? It will give you some practice, and also allow you to cover all the scenarios.

jthrilly (changes requested):

schemas/participant.ts
buckhalt: Thanks for catching this. I moved `trim()` after min checks but before `refine`. This simplifies the `hasNonWhitespaceCharacters` utility function, since the input is already trimmed.

I added extensive tests to cover all the scenarios for each schema, rows, and lists of rows. All passing. 

Refactored `participantIdentifierOptionalSchema` and `participantLabel` schema to use `trim() `and to not use the `hasNonWhitespaceCharacters`. Since these are optional, they should be trimmed and pass with whitespace or empty strings. These are then entered into the db either as undefined for label or assigned an id for identifier.

[7651da4](https://github.com/complexdatacollective/Fresco/pull/140/commits/7651da470f9b03c789d5b0fd680a90872bfda309) & [e1a0507](https://github.com/complexdatacollective/Fresco/pull/140/commits/e1a0507cdc653ca84c79c8ded34645042c50e3ae)

PR #150, opened 2024-06-23
PWA

jthrilly:
The idea here is to add PWA functionality to Fresco so that it can be "installed" on an iPad/tablet, and can run the interview experience fullscreen. 

This will give similar functionality to Interviewer, with the drawback of requiring an internet connection (unless we figure out something _very_ elaborate with caching).


jthrilly:
From Caden:

- [ ] Went through the sample protocol and was able to complete the entire interview with no major issues. there are a few ux things that are a little weird with varying levels of importance:
- [ ] scrolling on forms is slow and unresponsive. scrolling otherwise is fine. I will take a video to demonstrate this
- [ ] name generator - nodes can be nominated by clicking return on the screen keyboard. but you can’t see them once they’re added because you’d need to scroll up
- [ ] when the keyboard is up, it’s a little difficult to drag nodes from the side panel into the network. particularly on large roster interfaces where the user can search for something using the keyboard. when you “pick up” the node to start dragging it, you can’t see it anymore. it’s still possible to drop it in but is not intuitive. to see it, you need to either scroll up with the keyboard still up, or hide the keyboard. will take a video for this too.
- [ ] general navigation - when it is entirely in fullscreen, researcher navigation is tricky. we don’t have the ability to use the url. I can’t find an easy way to paste an interview url within the homescreen app. I  could only enter an in-progress interview. Also, once you get to the finished interview screen, i’m not sure how you would get back for interviewing others.
- [ ] scrolling forms is laggy (could also be true in the browser - needs testing)
- [ ] can't see nodes added to network when keyboard is up
- [ ] cant easily see or drag out nodes from the roster
- [ ] possible bug with navigating backwards to a name generator (sample protocol).

jthrilly:
Closing as no bandwidth to finish implementation, for now.

PR #154, opened 2024-07-08
Fix docker deployment related issues

mrkarimoff:
This PR fixes the following issues related to docker deployment:

- Fix uploading protocol by setting `config.callbackUrl` for UploadThing router if the app is running in production & `PUBLIC_URL` environment is set to the domain name of the app
- Fix the anonymous participation URL feature by setting the `url` to `new URL(env.PUBLIC_URL)` in `/onboard/protocolId` route if present

### DIFF
--- a/app/(interview)/onboard/[protocolId]/route.ts
+++ b/app/(interview)/onboard/[protocolId]/route.ts
@@ -1,6 +1,7 @@
 import { cookies } from 'next/headers';
 import { NextResponse, type NextRequest } from 'next/server';
 import { createInterview } from '~/actions/interviews';
+import { env } from '~/env';
 import trackEvent from '~/lib/analytics';
 import { getLimitInterviewsStatus } from '~/queries/appSettings';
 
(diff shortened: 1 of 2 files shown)


app/(interview)/onboard/[protocolId]/route.ts
```diff
@@ -11,7 +12,7 @@ const handler = async (
   { params }: { params: { protocolId: string } },
 ) => {
   const protocolId = params.protocolId; // From route segment
-  const url = req.nextUrl.clone();
+  const url = env.PUBLIC_URL ? new URL(env.PUBLIC_URL) : req.nextUrl.clone();
```
Comment (jthrilly) [inline app/(interview)/onboard/[protocolId]/route.ts:15]: Needs a comment to explain it, and perhaps better written with optional chaining rather then ternary.

app/api/uploadthing/route.ts
```diff
@@ -1,7 +1,14 @@
 import { createRouteHandler } from 'uploadthing/next';
+import { env } from '~/env';
 import { ourFileRouter } from './core';
 
 // Export routes for Next App Router
 export const { GET, POST } = createRouteHandler({
   router: ourFileRouter,
+  config: {
+    callbackUrl:
+      env.NODE_ENV === 'production' && env.PUBLIC_URL
```
jthrilly, line 10: Why is the check for production environment needed here (and if it is needed, why isn't it also used in the route change above)?

Again, would rather see nullish coalescing than ternary here.

jthrilly (changes requested):

jthrilly (approved):

PR #161, opened 2024-10-02
Feature/ Env variables in Db

buckhalt:
This feature moves several env variables to the db in the `appSettings` entity. This is primarily to facilitate an easier deployment process. 

- Ability to configure `UPLOADTHING_TOKEN` and `disableAnalytics` from the setup flow
- Ability to add an `UPLOADTHING_TOKEN` to existing deployments, via a modal
- Refactored appSettings to be a key/value pair
- Updated to UploadThing v7, which introduced environment variable changes and other breaking changes 
- Handles migrations


actions/appSettings.ts
```diff
@@ -1,58 +1,81 @@
...
+    },
   });
 
-  safeRevalidateTag('allowAnonymousRecruitment');
+  // handle revalidation tag
+  // safeRevalidateTag('appSettings', key);
+  return value;
+}
+
+export async function createAppSetting<
```
jthrilly, line 34: I'm not sure we need createAppSetting. Why can't setting just create if a record doesn't exist?

actions/appSettings.ts
```diff
@@ -1,58 +1,81 @@
...
 import { prisma } from '~/utils/db';
 
-export async function setAnonymousRecruitment(input: boolean) {
-  await requireApiAuth();
+// Generic function to set an app setting with validation
+export async function setAppSetting<
+  Key extends keyof z.infer<typeof appSettingSchema>,
+>(key: Key, value: z.infer<typeof appSettingSchema>[Key]) {
+  console.log('setting app setting', key, value);
+
```
jthrilly, line 16: I think this needs an auth check, right?

actions/appSettings.ts
```diff
@@ -1,58 +1,81 @@
...
+    data: {
+      key: key,
+      value:
+        typeof value === 'boolean' ? value.toString() : JSON.stringify(value),
+    },
   });
 
-  safeRevalidateTag('allowAnonymousRecruitment');
+  // handle revalidation tag
+  // safeRevalidateTag('appSettings', key);
```
jthrilly, line 30: It should be possible to make this work with the NextJS cache, by dynamically constructing a key based on the setting name (`appSettings-${key}`) and then revalidating in the setter, and wrapping the getting in `unstable_cache`.

queries/appSettings.ts
```diff
@@ -1,81 +1,61 @@
...
+>(key: Key): Promise<z.infer<typeof appSettingSchema>[Key]> {
+  const initializedAt = await prisma.appSettings.findUnique({
+    where: { key: 'initializedAt' },
+  });
 
-const calculateIsExpired = (configured: boolean, initializedAt: Date) =>
-  !configured && initializedAt.getTime() < Date.now() - UNCONFIGURED_TIMEOUT;
+  if (!initializedAt) {
+    // initialize the app
+    // todo: should these come from fresco.config?
```
jthrilly, line 21: Yes, and then I would make a separate function for `initialiseWithDefaults` or something.

jthrilly: You also should think about putting this in a transaction to ensure that they all succeed, or are rolled back. 

If you can't work out how to do that using `createAppSetting`, I think its okay to have the `initialiseWithDefaults` bulk create records with a one-off query.

queries/appSettings.ts
```diff
@@ -1,81 +1,61 @@
...
+  if (!keyValue) {
+    return null;
   }
 
-  const appSettings = appSettingsRecords.reduce((acc, { key, value }) => {
-    acc[key as AppSettingsKeys] = value;
-    return acc;
-  }, {} as AppSettings);
+  // Parse the value based on the expected type
+  const parsedValue =
```
jthrilly, line 38: I would expect you to be using either `preprocess()` or `coerce()` here. Zod normally wants to parse then validate, but because we already know the values coming from the db will be the wrong type for anything but strings, so you need to use those methods.

Maybe you should try writing some validation tests to make sure you get this working correctly?

actions/appSettings.ts
buckhalt: resolved [b9da84f](https://github.com/complexdatacollective/Fresco/pull/161/commits/b9da84f06fc178d6c9906a2aabf7e3958146c5d7)

actions/appSettings.ts
buckhalt: removed and consolidated with setAppSetting using upsert [b9da84f](https://github.com/complexdatacollective/Fresco/pull/161/commits/b9da84f06fc178d6c9906a2aabf7e3958146c5d7)

actions/appSettings.ts
buckhalt: resolved using existing safeRevalidateTag setup and `appSettings-${key}` [b9da84f](https://github.com/complexdatacollective/Fresco/pull/161/commits/b9da84f06fc178d6c9906a2aabf7e3958146c5d7)

queries/appSettings.ts
buckhalt: created initializeWithDefaults as a one-off query in a transaction [b9da84f](https://github.com/complexdatacollective/Fresco/pull/161/commits/b9da84f06fc178d6c9906a2aabf7e3958146c5d7). Note that I had to increase the timeout from the default bc this transaction takes longer than the default 5 seconds. Need to think about if this will cause problems with the vercel free tier like we've run into in the past.

queries/appSettings.ts
buckhalt: implemented using `preprocess()` and added validation tests [24c2a27](https://github.com/complexdatacollective/Fresco/pull/161/commits/24c2a27b6246ff76bd611535579698ad77b7f324)

actions/appSettings.ts
```diff
@@ -117,12 +106,32 @@ export async function storeEnvironment(formData: unknown) {
...
+                : JSON.stringify(value),
+          },
+        });
+      }
+    },
+    {
+      timeout: 10000, // default is 5000 which is too short for this
+    },
+  );
+}
```
jthrilly: Just thinking about this again....is there a reason not to use `createMany` in this case? Now we aren't calling separate functions, we don't need the transaction, right?

fresco.config.ts
```diff
@@ -11,3 +13,11 @@ export const dateOptions: Intl.DateTimeFormatOptions = {
   hour: 'numeric',
   minute: 'numeric',
 };
+
+export const DEFAULT_APP_SETTINGS = {
+  initializedAt: new Date(),
```
jthrilly, line 18: Should `initializedAt` and `installationId` have default values here? We probably want `installationId` to still be configurable via an env variable (maybe something like: if the env variable exists, that value is used when initialising, and it becomes "read only" in the ui?), and `initializedAt` feels more like it should be calculated as the db insert is happening. What do you think?

queries/appSettings.ts
```diff
@@ -17,13 +16,7 @@ export async function getAppSetting<
...
 
   if (!initializedAt) {
-    // initialize the app
-    // todo: should these come from fresco.config?
-    await createAppSetting('initializedAt', new Date());
-    await createAppSetting('configured', false);
-    await createAppSetting('allowAnonymousRecruitment', false);
-    await createAppSetting('limitInterviews', true);
-    await createAppSetting('installationId', createId());
+    await initializeWithDefaults();
```
jthrilly, line 18: You could make this return the values it has just inserted, and then return the value here from this function, rather than then doing another query to fetch the value after you just set it.

queries/appSettings.ts
```diff
@@ -87,12 +80,12 @@ export const getAnonymousRecruitmentStatus = createCachedFunction(async () => {
     'allowAnonymousRecruitment',
   );
   return getAnonymousRecruitmentStatus;
-}, ['allowAnonymousRecruitment', 'appSettings']);
+}, ['appSettings-allowAnonymousRecruitment', 'appSettings']);
```
jthrilly, line 77: Sorry for not explaining this well - I was actually suggesting we should _remove_ these specific actions/queries and just use the generic get/set settings functions, since those can automatically set cache keys and call revalidate.

queries/appSettings.ts
```diff
@@ -102,24 +95,24 @@ export const getUploadthingVariables = createCachedFunction(async () => {
     UPLOADTHING_SECRET,
     UPLOADTHING_APP_ID,
```
jthrilly, line 90: I don't think the v7 API actually uses the App ID (could be wrong).

Also, we don't have to keep the capitalisation from the env variables. Might be better to use pascalCase for consistency.

actions/appSettings.ts
buckhalt: resolved [5739d8d](https://github.com/complexdatacollective/Fresco/pull/161/commits/5739d8d788ef202113dca7f45afbaf157db5ba79)

queries/appSettings.ts
buckhalt: resolved [5739d8d](https://github.com/complexdatacollective/Fresco/pull/161/commits/5739d8d788ef202113dca7f45afbaf157db5ba79)

app/(blobs)/(setup)/_components/UploadThingTokenForm.tsx
```diff
@@ -21,7 +22,7 @@ export const UploadThingTokenForm = () => {
   }: {
     uploadThingToken: string;
   }) => {
-    await submitUploadThingForm({ uploadThingToken });
+    await action(uploadThingToken);
```
jthrilly, line 25: I think part of why we're not getting good UX is that we aren't using the built in form state management hooks for server actions. A good future improvement would be to create the `<Form>` component that we discussed, which would handle this functionality automatically.

jthrilly (approved):

schemas/environment.ts
jthrilly, line 1: This should be part of the `appSettings` schema.

schemas/environment.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from 'zod';
+import { zfd } from 'zod-form-data';
+
+// Custom parser for UPLOADTHING_TOKEN to remove token name and quotes
+const parseUploadThingToken = (token: string) => {
+  return token.replace(/^(UPLOADTHING_TOKEN=)?['"]?|['"]$/g, '').trim();
+};
+
+export const createUploadThingTokenForm = z.object({
```
jthrilly, line 9: Although this _is_ used in a form context, it doesn't use formData (which the schema below does) so the naming is confusing.

This should be `createUploadThingTokenSchema`.

app/dashboard/_components/UpdateUploadThing.tsx
jthrilly, line 1: This should be split out into a generic component that can update any (string) setting, so that it can be used for the installation ID, too.

app/dashboard/_components/UploadThingModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <DialogContent aria-describedby={undefined} className="sm:max-w-[500px]">
+        <DialogHeader>
+          <DialogTitle>Required Environment Variable Update</DialogTitle>
+          <DialogDescription>
+            <Paragraph>
+              This Fresco update requires a new UploadThing environment
+              variable. Please add it by following these steps.
+            </Paragraph>
+
+            <Paragraph>
```
jthrilly, line 27: Lists aren't allowed inside of paragraphs (isn't valid HTML).

app/dashboard/_components/UploadThingModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+                  <Link
+                    href="https://uploadthing.com/dashboard/"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                  >
+                    UploadThing dashboard
+                  </Link>
+                </li>
+                <li>Select your project.</li>
+                <li>Select the API Keys tab.</li>
```
jthrilly, line 40: I think it would be useful to have an image here, since this is a UI change on the UT side too, and there are some confusing elements to it.

app/dashboard/_components/UploadThingModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  DialogDescription,
+  DialogHeader,
+  DialogTitle,
+} from '~/components/ui/dialog';
+import Paragraph from '~/components/ui/typography/Paragraph';
+
+function UploadThingModal() {
+  return (
+    <Dialog open={true}>
+      <DialogContent aria-describedby={undefined} className="sm:max-w-[500px]">
```
jthrilly, line 18: We shouldn't ever really override the styling of such a fundamental component like this. If the width is wrong here, it should (in principle) be wrong everywhere, and so should be updated in the component itself. 

Also, if you _do_ need to make a one off adjustment, you shouldn't use arbitrary values for the same reason (prefer updating the breakpoints if those values aren't correct).

app/dashboard/_components/UploadThingModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  DialogContent,
+  DialogDescription,
+  DialogHeader,
+  DialogTitle,
+} from '~/components/ui/dialog';
+import Paragraph from '~/components/ui/typography/Paragraph';
+
+function UploadThingModal() {
+  return (
+    <Dialog open={true}>
```
jthrilly, line 17: Forcing this to open means that the user can't use the app until they fix this issue. While it is important that it be fixed, the app can still be used without adding the token. This should be closable.

app/dashboard/settings/page.tsx
```diff
@@ -28,7 +37,25 @@ export default async function Settings() {
...
+
+          {!!env.INSTALLATION_ID && (
+            <Alert variant="info" className="mt-4">
+              <AlertTitle>Note:</AlertTitle>
+              <AlertDescription>
+                This setting is controlled by your <code>.env</code> file, and
+                so can only be changed by modifying that file.
+              </AlertDescription>
+            </Alert>
+          )}
```
jthrilly, line 55: This should be refactored into a reusable component

app/dashboard/settings/page.tsx
```diff
@@ -28,7 +37,25 @@ export default async function Settings() {
...
+          <Input disabled={!!env.INSTALLATION_ID} value={installationId} />
```
jthrilly, line 56: It should be possible to share the update functionality for this with the uploadthing editable input

app/dashboard/settings/page.tsx
```diff
@@ -60,8 +87,31 @@ export default async function Settings() {
...
+          </Paragraph>
+          {!!env.DISABLE_ANALYTICS && (
+            <Alert variant="info" className="mt-4">
+              <AlertTitle>Note:</AlertTitle>
+              <AlertDescription>
+                This setting is controlled by your <code>.env</code> file, and
+                so can only be changed by modifying that file.
+              </AlertDescription>
+            </Alert>
+          )}
```
jthrilly, line 110: Should use the shared component

PR #162, opened 2024-10-21
Potential changes

jthrilly:


queries/appSettings.ts
```diff
@@ -83,21 +44,22 @@ export async function requireAppNotExpired(isSetupRoute = false) {
...
-  const configured = await getAppSetting('configured');
+  const isConfigured = await getAppSetting('configured');
   const initializedAt = await getAppSetting('initializedAt');
 
-  if (!configured || !initializedAt) {
-    throw new Error('Could not get app settings');
-  }
-  return calculateIsExpired(configured, initializedAt);
+  return (
+    !isConfigured && initializedAt.getTime() < Date.now() - UNCONFIGURED_TIMEOUT
```
buckhalt, line 61: This is throwing an error on resetting the db and on changing setup steps that `initializedAt.getTime()` is not a function. I think that there's a problem in the way getAppSetting is preprocessing the value as a date.

app/(blobs)/(setup)/_components/UploadThingTokenForm.tsx
```diff
@@ -40,6 +43,7 @@ export const UploadThingTokenForm = () => {
         />
       </div>
       <div className="flex flex-wrap">
+        <SubmitButton>Submit</SubmitButton>
         <Button disabled={isSubmitting || !isValid} type="submit">
```
buckhalt, line 47: Previous button implementation needs to be removed

app/dashboard/settings/page.tsx
```diff
@@ -28,7 +33,31 @@ export default async function Settings() {
...
+              <AlertTitle>Note:</AlertTitle>
+              <AlertDescription>
+                Your installation ID was set using your <code>.env</code> file,
+                and so can only be changed by modifying that file.
+              </AlertDescription>
+            </Alert>
+          )}
+          <Input disabled={envInstallationId} value={installationId} />
+        </SettingsSection>
+        <SettingsSection heading="UploadThing API Key">
```
buckhalt, line 54: Is this intended to be read only or editable? I think the input either needs to be disabled like the installationId, or editing functionality should be implemented

app/(blobs)/(setup)/_components/OnboardSteps/Documentation.tsx
```diff
@@ -1,18 +1,28 @@
...
 import SubmitButton from '~/components/ui/SubmitButton';
 import Heading from '~/components/ui/typography/Heading';
 import Paragraph from '~/components/ui/typography/Paragraph';
+import { env } from '~/env';
 import trackEvent from '~/lib/analytics';
 
 function Documentation() {
   const handleAppConfigured = async () => {
-    await setAppConfigured();
+    const installationId = env.INSTALLATION_ID ?? createId();
```
buckhalt, line 15: This is returning undefined even with an installationId in the env. Needs investigation.

buckhalt (changes requested):
Overall this is looking great. Particularly like the simplification of `setAppSetting` and the setup flow, as well as the settings page improvements. I found a few small issues above that I'm happy to work on.

app/dashboard/settings/page.tsx
buckhalt: Marked the input as `readonly` to resolve the error here [baefaab](https://github.com/complexdatacollective/Fresco/pull/162/commits/baefaab812a4b0195f2d9c30c665cd2beceda3bc). Can be removed if editing functionality is added

app/(blobs)/(setup)/_components/SignUpForm.tsx
```diff
@@ -18,6 +19,7 @@ export const SignUpForm = () => {
 
   const onSubmit = async (data: unknown) => {
     await signup(data);
+    redirect('/setup?step=2');
```
buckhalt: Using redirect here throws a runtime error. Next [docs](https://nextjs.org/docs/app/building-your-application/routing/redirecting#userouter-hook) recommend using `useRouter` and `router.push()` in client components. Any reason not to here?

queries/appSettings.ts
buckhalt: Looked into this more and it is being correctly preprocessed. The only way I can get this to work is to explicitly create a new `Date` in this function before using `getTime`. Implemented here [9cca019](https://github.com/complexdatacollective/Fresco/pull/162/commits/9cca019e3b866a31ee57b830f2abdd5e67bef01e)

app/(blobs)/(setup)/_components/SignUpForm.tsx
buckhalt: potential approach with `useRouter` instead here [9cca019](https://github.com/complexdatacollective/Fresco/pull/162/commits/9cca019e3b866a31ee57b830f2abdd5e67bef01e)

PR #163, opened 2024-10-24
Feature: upgrade notification system

buckhalt:
Feature to provide more detailed update notifications to users in the settings page. Uses release to compare versions, provide generic info for patch/minor/major updates, and render release notes.

### DIFF
--- a/components/VersionSection.tsx
+++ b/components/VersionSection.tsx
@@ -1,208 +1,214 @@
-'use client';
-
-import { CheckCircle2, Info, Loader2, XCircle } from 'lucide-react';
-import { use, useEffect, useState } from 'react';
+import { AlertTriangle, CheckCircle2, Info, Loader2 } from 'lucide-react';
+import { unstable_noStore } from 'next/cache';
 import Markdown from 'react-markdown';
 import { z } from 'zod';
 import Link from '~/components/Link';
 import { Alert, AlertDescription, AlertTitle } from '~/components/ui/Alert';
 import { env } from '~/env';
-import { type getInstallationId } from '~/queries/appSettings';
+import trackEvent from '~/lib/analytics';
+import { getInstallationId } from '~/queries/appSettings';
 import { ensureError } from '~/utils/ensureError';
 import SettingsSection from './layout/SettingsSection';
+import { Button } from './ui/Button';
+import Heading from './ui/typography/Heading';
 import Paragraph from './ui/typography/Paragraph';
 
-type SemVerUpdateType = 'major' | 'minor' | 'patch';
+const semverSchema = z
+  .string()
+  .regex(
+    /^v(\d+)\.(\d+)\.(\d+)$/,
+    "Invalid version format. Expected format is 'v1.2.3'.",
+  )
+  .transform((version) => {
+    const [, major, minor, patch] = version.match(
+      /^v(\d+)\.(\d+)\.(\d+)$/,
+    ) as string[];
 
-const GithubApiResponseSchema = z.object({
-  tag_name: z.string().transform((tag) => tag.replace('v', '')), // remove the v to compare with APP_VERSION
-  body: z.string(),
-});
+    if (!major || !minor || !patch) {
+      throw new Error('Invalid version format');
+    }
 
-function getSemverUpdateType(
-  currentVersion: string,
-  mainVersion: string,
-): SemVerUpdateType | null {
-  if (currentVersion === mainVersion) return null;
+    // Convert version parts to numbers
+    const majorNum = parseInt(major, 10);
+    const minorNum = parseInt(minor, 10);
+    const patchNum = parseInt(patch, 10);
 
-  const [currentMajor, currentMinor, currentPatch] = currentVersion.split('.');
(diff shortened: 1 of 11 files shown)


components/VersionSection.tsx
```diff
@@ -2,19 +2,39 @@
...
 import SettingsSection from './layout/SettingsSection';
 import Paragraph from './ui/typography/Paragraph';
 
+type SemVerUpdateType = 'major' | 'minor' | 'patch';
+
 const GithubApiResponseSchema = z.object({
-  status: z.string(),
-  ahead_by: z.number(),
-  behind_by: z.number(),
+  tag_name: z.string().transform((tag) => tag.replace('v', '')), // remove the v to compare with APP_VERSION
```
jthrilly, line 18: I think it would be better to do this parsing in a RegExp, which could also extract the elements of the version.

While we are at it, why not use zod to parse this into an _actual_ semver? So we give it a string, and we get out an object with major, minor, and patch properties?

components/VersionSection.tsx
```diff
@@ -2,19 +2,39 @@
...
+  mainVersion: string,
+): SemVerUpdateType | null {
+  if (currentVersion === mainVersion) return null;
+
+  const [currentMajor, currentMinor, currentPatch] = currentVersion.split('.');
+  const [mainMajor, mainMinor, mainPatch] = mainVersion.split('.');
+
+  if (mainMajor !== currentMajor) return 'major';
+  if (mainMinor !== currentMinor) return 'minor';
+  if (mainPatch !== currentPatch) return 'patch';
```
jthrilly, line 33: You're comparing these as strings, and only checking if they are different. It is a fairly reasonable assumption, but versions are actually numbers. If you do "1" > "2", it will actually work, but it isn't intentional. 

I think a better way would be to use the zod schema to cast these as numbers, and then to do `currentVersion < newVersion`.

components/VersionSection.tsx
```diff
@@ -56,61 +104,105 @@ export default function VersionSection({
...
+      ) : (
+        <>
+          <CheckCircle2 className="h-8 w-8 fill-success text-success-foreground" />
+          <Paragraph variant="smallText" margin="none">
+            You are up to date!
+          </Paragraph>
+        </>
+      )}
+    </div>
+  );
```
jthrilly, line 136: I wouldn't make this a separate closure - I would have the separate "states" in the render function, and use early return and tighter conditions rather than in a big ternary.

components/VersionSection.tsx
```diff
@@ -56,61 +104,105 @@ export default function VersionSection({
...
+
+      {data?.semVerUpdateType && (
+        <>
+          <Alert
+            variant={data.semVerUpdateType === 'major' ? 'default' : 'info'}
+            className="mt-4"
+          >
+            <AlertTitle>
+              {' '}
+              <div className="mt-4 flex flex-row items-center space-x-2">
```
jthrilly, line 157: Hmm. This isn't the correct usage. If you check the other alerts you'll see that the icon goes outside the title element (just before it, actually), and that this triggers adding padding to the left automatically.

components/VersionSection.tsx
```diff
@@ -56,61 +104,105 @@ export default function VersionSection({
...
+                <Info className="h-8 w-8 text-info" />
+                <Paragraph
+                  className="text-info"
+                  variant="smallText"
+                  margin="none"
+                >
+                  A new version of Fresco is available: {data.newVersion}
+                </Paragraph>
```
jthrilly, line 165: This definitely shouldn't be a <Paragraph>! (1) It is already inside an `Alert.Title` (which renders a heading) and (2) you've styled a paragraph like a heading!

components/VersionSection.tsx
```diff
@@ -56,61 +104,105 @@ export default function VersionSection({
...
+                  </Alert>
+                )}
+              </Paragraph>
+
+              <Paragraph>
+                <article className="prose prose-a:text-blue-600 prose-headings:text-sm prose-headings:text-foreground prose-headings:font-extrabold prose-headings:foreground prose-headings:tracking-widest prose-headings:uppercase text-sm text-foreground">
+                  <Markdown>{data.releaseNotes}</Markdown>
+                </article>
+              </Paragraph>
+            </AlertDescription>
```
jthrilly, line 202: Also issues here with the structure:

- AlertDescription renders a div with the paragraph variants. There's no need for nested paragraphs inside. If you need nested paragraphs - don't use the AlertDescription.
- As per the radix docs, the description is important for screen readers, so I would keep it to just the basic info (first paragraph). However, that paragraph doesn't actually directly say that there's a new version available!
- The nested alert should _not_ be wrapped inside a Paragraph! It is a block level element that renders a div, and divs inside of p is invalid HTML. A DIV is conceptually meant to represent a container, which has its own spacing and alignment relative to other elements. 
- You also conditionally render the Alert, but not the paragraph. This will lead to empty markup, which would be bad for screen readers, and also potentially will add additional spacing due to margin/padding.
- The same goes for putting an `<article>` inside of a paragraph - never do this. Article is a more semantic form of div, and is block level. 
- You used an invalid color there (blue-600).

jthrilly (commented):
A few issues here, but coming along.

PR #166, opened 2024-11-01
Fix/fresh setup migrations

jthrilly:
This PR is an attempt to deal with some issues relating to fresh installs. 

Essentially the migration process we have now will fail for someone deploying the app for the first time with v2.0.0. This is because the database schema will be the 2.0.0 schema (with the changes to AppSettings), but _without_ the migrations table. This in turn will trigger the migrations to run, including the 2.0.0 migration, which will fail because the table structure already exists.

The solution I have come up with is to use `prisma migrate diff` to calculate if the current database schema differs from the schema file, and to use this to determine if `prisma migrate dev` should be run. This seems to work.

I had to separate out the seeding of the initializedAt key, because the underlying database schema might have changed _after_ the prisma client was initialized. I could also have dynamically imported it a second time, perhaps.

Todo: test _all_ update scenarios thoroughly!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ WORKDIR /app
 COPY prisma ./prisma
 
 # Copy package.json and lockfile, along with postinstall script
-COPY package.json pnpm-lock.yaml* postinstall.js migrate-and-start.sh setup-database.js ./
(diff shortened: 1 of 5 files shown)


jthrilly (commented):
Going to merge, and test in the context of the next branch

PR #167, opened 2024-11-01
Fix: Semantic versioning logic

buckhalt:
Fixes bug where incorrect update type was returned when current version is higher than new version. Adds tests to cover logic.

Todo:
- [ ] bump version and generate release notes


jthrilly (approved):
Really excellent. Perfect resolution, and love that you included test coverage.

PR #204, opened 2024-11-11
Implement DOMBuilder approach to XML generation

jthrilly:
The idea here is to avoid the risky and potentially error prone string manipulation technique we use to generate XML, and to instead use the dom builder directly to create and manipulate elements and attributes.

The downside of this approach is that it is incompatible with streaming the result to disk.

WIP. Functioning, but without thorough testing.


jthrilly:
Superseded by https://github.com/complexdatacollective/Fresco/pull/318

PR #228, opened 2024-11-29
Schema 8

jthrilly:
This PR provides a staging area for the new interfaces we are developing for schema 8.

Also contained here:
- Update to redux and redux toolkit
- inlining of protocol validation lib
- removal of device settings redux slice


.eslintrc.cjs
```diff
@@ -26,11 +26,18 @@ const config = {
...
     'prettier',
   ],
-  ignorePatterns: ['node_modules', '*.stories.*', '*.test.*', 'public', '.eslintrc.cjs',],
+  ignorePatterns: [
+    'node_modules',
+    '*.stories.*',
+    '*.test.*',
+    'public',
+    '.eslintrc.cjs',
+    'lib/protocol-validation', // TODO: remove this, and fix the errors
```
jthrilly, line 35: Do not merge without removing this

lib/protocol-validation/__tests__/test-protocols.test.ts
jthrilly, line 1: This needs to be implemented or removed before merging

PR #236, opened 2024-12-06
Geospatial interface

buckhalt:
Implements the Geospatial Interface for assigning node attributes by selecting areas on a map, using mapbox


utils/protocolImport.tsx
```diff
@@ -21,14 +21,37 @@ export const getProtocolJson = async (protocolZip: Zip) => {
...
+type FileAsset = {
+  file: File;
+} & ProtocolAsset
+
+type ApiKeyAsset = {
+  value: string;
+  key: string;
+  size: number;
+  url: string;
+} & ProtocolAsset
```
jthrilly, line 40: Aren't these types defined anywhere else? It feels like these should live in either the protocol schema, the prisma schema, or shared-consts.

buckhalt: resolved [28ede8c](https://github.com/complexdatacollective/Fresco/pull/236/commits/28ede8cd31511e81182a96c7c536f081054252d7)

PR #253, opened 2025-01-13
Resolve animation issues after Tailwind upgrade

jthrilly:
Resolve issue where leading 0 was being stripped from CSS variables in tailwind v4 alpha.

- I used parseInt for coercing to numbers instead of Number() or parseFloat. The latter two would have handled strings without a leading zero without an issue.
- Tailwind was reserializing all CSS files, even when they didn't use any tailwind directives. This was presumably causing the zeros to be stripped. This was fixed in beta.5


lib/interviewer/hooks/useAnimationSettings.js
jthrilly, line 1: I think we should delete this file. useAnimation settings is only used in two places, and both places are used in framer motion animations which can just use the css variable directly

lib/ui/components/Transitions/Fade.js
```diff
@@ -19,7 +18,7 @@ function Fade({
     exit: getCSSVariableAsNumber('--animation-duration-fast-ms'),
   };
 
-  const defaultEasing = getCSSVariableAsObject('--animation-easing-js');
+  const defaultEasing = '--animation-easing-js';
```
jthrilly, line 21: This won't work here, I don't think? Did you test it? I'm not sure where you got the idea that animejs accepts a css variable for the easing parameter, but [it doesn't seem like it does](https://animejs.com/documentation/#staggeringEasing).

What I suggest we do is to replace this with `cubicBezier(0.4, 0, 0.2, 1)` (which should correspond to the CSS easing value). Yes it is duplication. No, it isnt DRY. Its fine here.

Ultimately, we want to delete this file and remove animejs as a dependency, too.

lib/interviewer/hooks/useAnimationSettings.js
buckhalt: Removed `useAnimationSettings` and replaced with using css variables directly, through the `style` prop.

There was one instance of doing an additional calculation on duration which requires it to be a number. I used `getCSSVariableAsNumber` directly there:https://github.com/complexdatacollective/Fresco/blob/ef0a4c3704bb66cd272eca4c1b2b2b1416c981d4/lib/interviewer/components/List.js#L94.

buckhalt:
Tested broken features on prod & dev using the Sample Protocol. All comments resolved. Merging this in!

PR #299, opened 2025-02-20
Fix geospatial ready for next

buckhalt:
Fixes logic to make readyForNext properly triggered when there is a response on the geospatial interface. Also improves copy on outside selectable areas overlay to be more clear.


buckhalt:
Checks are failing bc of the protocol-validation package stuff in the schema-8 branch. Unrelated to anything here.

PR #301, opened 2025-02-26
Next

jthrilly:


app/dashboard/_components/UserMenu.tsx
```diff
@@ -3,7 +3,7 @@ import SubmitButton from '~/components/ui/SubmitButton';
 
 const UserMenu = () => {
   return (
-    <form action={() => logout()}>
+    <form action={void logout}>
```
jthrilly, line 6: needed because the type signature of actions seems to have changed to forbid returning values. We don't use the returned value, so the action could also be refactored, but I left as is and just voided the call.

PR #349, opened 2025-04-07
Fix form skip logic

buckhalt:
#### Issue: 
Interview navigation functions (`moveForward` and `moveBackward`) were using stale values of `nextValidStageIndex` and `previousValidStageIndex` due to closures. When these indexes were updated during form submission in the `beforeNext` handler, the navigation functions weren't receiving the updated values. This resulted in behavior where if form data impacted skip logic, the skip logic wasn't working correctly.

#### Solution:
- Implemented using `useRef` to store index values as refs
- Added `useEffect` hooks to update these refs whenever indexes change
- Modified `moveForward` and `moveBackward` to use current ref values for updating session progress and stage instead of closure-captured variables

#### Other changes:
This investigation also revealed a circular deps issue and duplicate selectors (two duplicate versions of `getStageIndex`). Fixed this by moving some selectors (`getActiveSession`, `getStageIndex`) to a `shared` file and importing from there.

### DIFF
--- a/app/(interview)/interview/_components/ServerSync.tsx
+++ b/app/(interview)/interview/_components/ServerSync.tsx
@@ -5,7 +5,7 @@ import { type ReactNode, useCallback, useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import type { SyncInterviewType } from '~/actions/interviews';
 import usePrevious from '~/hooks/usePrevious';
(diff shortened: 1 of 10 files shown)


jthrilly (approved):
Looks great. Only thing for the future would be to remove some of the unnecessary comments, but nbd.

PR #439, opened 2025-07-29
Fix quick node form

buckhalt:


jthrilly (approved):
I think we will end up tweaking the styling a bit eventually, but this is a lovely fix in the meantime.

PR #718, opened 2026-04-20
fix(categorical-bin): size circles to content box when a bin is expanded

buckhalt:
When a Categorical Bin is expanded, other bins were sometimes overflowing the container. This was due to bins not resizing to fit the new content area after padding is added to reserve space for the expanded bin's panel.

Fix involves measuring the flex container's content box in `useCircleLayout` by subtracting padding from `clientWidth`/`clientHeight` so that the remaining bins are sized for the space actually available to them.

This made the closed bins sizes too small. Added logic to shrink the expanded panel as bin count grows.

**Before:**
Screenshot 2026-04-20 at 2 18 43 PM

**After subtracting padding from `clientWidth`/`clientHeight` (bins too small):**
Screenshot 2026-04-20 at 2 18 56 PM

**Final:**
Screenshot 2026-04-21 at 9 15 12 AM

Reproducible in the Categorical Bin storybook (Interview/Interfaces/CategoricalBin -> Default) with 10 bins using `categoryCount` and clicking a bin to expand it. 

todo:

- [x] update snapshots


jthrilly:
I think we need to add some logic to this for high bin numbers so that the open panel takes up less space. The bins are far too small right now.

PR #793, opened 2026-06-17
security: harden auth, storage, export and infra for 4.0.0

jthrilly:
## Summary

Implements remediations from the pre-release **security & correctness audit** of 4.0.0, in one PR. All changes are calibrated to Fresco's actual trust model (single-class admin; participants authorized by possession of an unguessable interview id), so the participant interview-taking flow stays unauthenticated by design.

**Verification:** `pnpm typecheck` ✓ · `pnpm lint` ✓ · `pnpm test` ✓ (344 unit tests, incl. new coverage below).

## High

- **Unauthenticated admin-account creation.** `signup` / `signupWithPasskey` / `generateSignupRegistrationOptions` now enforce the "app not configured" invariant *inside the action* (Server Actions are directly invokable, so the setup-page guard was insufficient). `completeSetup` now requires auth. New `isAppConfigured()` helper in `queries/appSettings.ts`.

## Medium

- **Asset object access (`/api/assets/[key]`).** Validates key format and only signs GETs for keys referenced by an `Asset` or a protocol's original file — it can no longer sign a GET for an arbitrary bucket object.
- **Stored XSS via uploaded assets.** Served with a validated `Content-Type`, `X-Content-Type-Options: nosniff`, and `Content-Disposition: attachment` for script-capable types (SVG/HTML). Inline `/<video>` rendering is unaffected.
- **CSV / formula injection.** New `escapeCsvFormula` util applied to the activity-feed export; the `@codaco/network-exporters` CSV serializer is patched (pnpm patch) to neutralize `= + - @ \t \r`-leading cells. The unauthenticated onboard `participantIdentifier` is now validated.
- **Secrets at rest.** API tokens are stored as a **SHA-256 hash** (plaintext shown once). Reversible storage secrets (S3/UploadThing) remain document-only per decision — SECURITY.md and `.env.example` now steer operators to env-var storage config to keep them out of the DB.
- **Interview data integrity.** Sync no longer trusts the client `lastUpdated` (Prisma `@updatedAt` sets it server-side), removing the dashboard sort/filter/export-ordering corruption vector; sync returns a generic 400 instead of the raw Zod error.

## Low / hardening

- **Security headers** added in `next.config.ts` (nosniff, `X-Frame-Options: SAMEORIGIN`, HSTS, `Referrer-Policy`), with **`no-referrer` on `/interview/*` and `/onboard/*`** so capability-bearing URLs don't leak via `Referer`. Interview ids are no longer logged.
- **SSRF:** admin-configured `s3Endpoint` is rejected if it targets a private/loopback/link-local address (`utils/safeUrl.ts`).
- **Login timing** equalized (dummy hash) to block username enumeration; failed login-attempt records are now awaited so the rate limiter counts them.
- **Dependencies:** `effect` override forces the transitive copy off the vulnerable `<3.20` (GHSA-38f7-945m-qr2g) onto the `3.21.3` we already use. *(Remaining `hono` advisory is under `@prisma/dev` — dev tooling, not the runtime path.)*
- **Infra:** MinIO root credentials are now required (no `minioadmin` default); removed the `.:/app/next-app` bind-mount that exposed `.env` into the app container (all 3 compose files); `/api/health` no longer discloses Node version / `NODE_ENV`; removed `typescript.ignoreBuildErrors` so type errors fail the build.
- Export `concurrency` is clamped; generated participant URLs now `encodeURIComponent` the identifier.

## Tests added

`utils/__tests__/escapeCsvFormula.test.ts`, `utils/__tests__/safeUrl.test.ts`, `actions/__tests__/apiTokens.test.ts` (hash-on-lookup), the signup-gate case in `actions/__tests__/auth.test.ts`, and expanded `app/api/assets/__tests__/route.test.ts` (DB scoping, key validation, svg disposition).

## Docs

SECURITY.md (security model + data-at-rest + headers), `.env.example` (storage + required MinIO vars), and stale `CLAUDE.md` paths fixed (`lib/db/schema.prisma`, `~/lib/db`, `@codaco/network-exporters`).

## Deferred (per audit, low risk)

- **Interview id entropy** (cuid v1 → CSPRNG token): kept cuid; closed the practical leak vectors (logging + `Referer`) and documented. A separate access-token column is the follow-up.
- **Session sliding refresh** (`active_expires`): documented the fixed-14-day behavior; left the schema unchanged.
- **Finish-route atomic final-network write**: low-severity ordering edge, left as-is.

> ⚠️ Since 4.0.0 has not shipped, the API-token hashing change needs no data migration (no production 4.0 tokens exist). The `network-exporters` patch is pinned to `1.0.3`; the durable fix is upstream.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


app/api/export-interviews/batch/route.ts
```diff
@@ -33,6 +33,21 @@ export async function POST(request: Request) {
...
+    typeof exportOptions.concurrency === 'number'
+      ? {
+          ...exportOptions,
+          concurrency: Math.min(
+            Math.max(1, Math.floor(exportOptions.concurrency)),
+            MAX_EXPORT_CONCURRENCY,
+          ),
+        }
+      : exportOptions;
+
```
jthrilly: MAX_EXPORT_CONCURRENCY shoudl be defined as part of the ExportOptionsSchema as a max value. This will cause a zod error if an inappropriate value is used.

app/dashboard/_components/ActivityFeed/ExportActivityFeed.tsx
```diff
@@ -20,8 +21,8 @@ export default function ExportActivityFeed() {
 
         const csvData = activities.map((activity) => ({
           timestamp: activity.timestamp.toISOString(),
-          type: activity.type,
-          details: activity.message,
+          type: escapeCsvFormula(activity.type),
+          details: escapeCsvFormula(activity.message),
```
jthrilly: Why do we need to escape potential CSV formula for data that we control and create? This seems paranoid.

patches/@codaco__network-exporters@1.0.3.patch
jthrilly, line 1: Please remove this patch. The escape sequence is overly cautious - we are encoding data that we generated and stored.

utils/escapeCsvFormula.ts
jthrilly, line 1: Candidate for deletion based on comments above.

SECURITY.md
```diff
@@ -4,6 +4,23 @@
...
 
+## Security Model
+
+Fresco is a **single-tenant, single-trust-level** application. Understanding this is important when reasoning about the codebase:
+
+- **Administrators** authenticate (password, with optional TOTP or passkey two-factor) and are all fully trusted — there is no role hierarchy. Anyone who can sign in can manage every protocol, participant, and interview. "Any logged-in user can do X" is therefore by design, not a privilege-escalation flaw.
+- **Participants** take interviews **without authenticating**. A participant is authorized purely by possession of their interview URL, which contains an unguessable interview id. **Treat interview URLs as secrets**: do not post them publicly, and Fresco does not write them to logs.
+- **Account creation** is only possible during initial setup. Once the app is configured, the signup actions reject all new accounts (the check is enforced inside the actions, not only on the setup page).
```
jthrilly: This is not correct. New accounts can be created from the settings page.

app/api/export-interviews/batch/route.ts
jthrilly: Done — moved the bound into `exportInterviewsSchema` (`schemas/export.ts`) as a refine on `exportOptions`, so an out-of-range `concurrency` now fails validation (400) instead of being silently clamped, and removed the clamp from the route. (5e47686)

app/dashboard/_components/ActivityFeed/ExportActivityFeed.tsx
jthrilly: Removed. Confirmed the activity-feed CSV cells never start with participant-controlled input: `type` is an app constant and every `message` is prefixed with literal text (`User …`, `Participant "…`), so a formula trigger can't land at position 0. Escaping here was unnecessary.

utils/escapeCsvFormula.ts
jthrilly: Deleted `utils/escapeCsvFormula.ts` and its test — it was only used by the activity-feed export above.

SECURITY.md
jthrilly: Fixed. The bullet now distinguishes the unauthenticated setup signup (gated to before configuration) from authenticated admins creating accounts via dashboard settings → User Management.

patches/@codaco__network-exporters@1.0.3.patch
jthrilly: Removed the activity-feed escaping above, but this one is a different case so I want to confirm before dropping it. The network-export CSV cells are node/edge **attribute values** — participant-typed interview responses (name generators, free-text prompts) — plus the participant identifier (which can arrive unauthenticated via `/onboard` under anonymous recruitment). Those are untrusted participant input, not values we generate, and can begin with `=`/`+`/`-`/`@`, so opening such an export in Excel/Sheets would evaluate a participant-supplied formula. Modern spreadsheets warn before executing and the API docs point consumers at R/Python, so accepting the risk is defensible — I just want to confirm given the data is participant-controlled. Happy to remove the patch if so.

jthrilly: Removed the patch (dd98ca7) — the package is back to upstream. The fix will be made directly in `@codaco/network-exporters` (`sanitizeCellValue` in the CSV formatter) so it covers every consumer rather than living as a Fresco-side patch.

PR #803, opened 2026-06-26
chore(deps): update @codaco packages and migrate collected categorica…

jthrilly:
…l data

Bump all @codaco packages to latest, including @codaco/protocol-utilities 1.x -> 2.0.0. Adapt the generateNetwork call sites in the test-interview generator to the new options-object signature (seed moved into options).

@codaco/interview 1.1.0 stores categorical attribute values as arrays of selected option values, dropping the scalar-handling fallbacks. Add a one-time, idempotent migration that wraps scalar categorical values in collected interview networks into single-element arrays, resolving categorical variables from the protocol codebook. Wired into the build:platform migration step after the v8 protocol migration.

Closes #795


jthrilly (commented):
I am concerned about the scenario where the upgrade from version 3.x of Fresco (using the old interview module) to version 4.x fails. If any of the steps in setup-database fail, we need to revert data migrations so as not to cause issues for the user.

